PR #150, opened 2026-06-05
#96/SpeedLimitLite blessed CSV baselines for VISSIM 2022 + 2026

yunlishao-vibe:
## Summary

Follow-up to merged #149. Adds blessed SpeedLimitLite output CSVs for both VISSIM 2022 and VISSIM 2026, plus the plumbing to run / compare against either:

- `tests/Vissim/SpeedLimitLite/speed_limit_lite_orig_v2022.csv` — 380 rows, produced by VISSIM 2022.00-13, ego id=15
- `tests/Vissim/SpeedLimitLite/speed_limit_lite_orig_v2026.csv` — 380 rows, produced by VISSIM 2026, ego id=13

`start_vissim.py` gains a `--progid` CLI option (or `VISSIM_PROGID` env var) so contributors on 2026-only dev boxes can run without editing the script. `compare_traces.py` gains a `--vissim-version {2022,2026}` flag that picks the matching reference CSV; `--ref` still works as an explicit override for comparing against the historical SpeedLimit Parquet references.

Both baselines verified self-consistent (0.0000 diff against themselves) and cross-version-divergent on every field (proof they're real different runs, not accidental duplicates).

CSV (not Parquet) because at this scale (~22 KB each) the columnar compression advantage doesn't matter, and CSV is git-diffable + readable in any text editor for PR review.

## Related Issues / Tasks

Continues the #96 / milestone 0.9.0 work. PR #149 landed the test infrastructure; this PR locks in the expected outputs.

## Type of Change

- [x] Test case / scenario update
- [x] New feature (the `--vissim-version` / `--progid` switches)

## Affected Modules / Components

- `tests/Vissim/SpeedLimitLite/start_vissim.py` — `--progid` / `VISSIM_PROGID` env var
- `tests/Vissim/SpeedLimitLite/compare_traces.py` — `--vissim-version {2022,2026}` picks matching CSV; default mod path switched to `.csv`
- `tests/Vissim/SpeedLimitLite/speed_limit_lite_orig_v2022.csv` — new
- `tests/Vissim/SpeedLimitLite/speed_limit_lite_orig_v2026.csv` — new
- `tests/Vissim/SpeedLimitLite/README.md` — documents the two-baseline story + override workflow

## Test Cases

Default contributor workflow:
```bat
cd tests\Vissim\SpeedLimitLite
run_speed_limit_lite.bat
python compare_traces.py             # vs v2022 baseline
python compare_traces.py --vissim-version 2026
```

VISSIM-2026-only dev box:
[...]

Issue #147, opened 2026-06-04
[MAINTENANCE] Rename VISSIM driver model DLLs: promote int-API build to default, freeze long-API build as legacy


yunlishao:
**Task Summary**
Rename the two VISSIM driver model DLLs so the int-API (v2021+) build becomes the unmarked default `DriverModel_RealSim.dll`, and the existing long-API build is frozen as `DriverModel_RealSim_legacy.dll`. The legacy DLL receives no future maintenance — it is kept solely as a compatibility shim for users on VISSIM ≤ 2020.

**Why is this needed?**
Follow-up to #129 (which consolidated both driver model variants behind a shared `DriverModel_FIXS_Common.h`). FIXS officially supports VISSIM 2022, which uses PTV's int-based driver-model API. The int-API build is therefore the canonical one and should carry the unmarked name; the long-API build remains only for the diminishing set of users still on VISSIM ≤ 2020.

Goals:
- Eliminate the confusing `_v2021` suffix on the actually-current DLL.
- Make it obvious which DLL is the supported default and which is legacy.
- Reduce ongoing maintenance burden by declaring the legacy DLL frozen (changes to it are not accepted).

**Scope of Work**

*Rename within `ProprietaryFiles` (proprietary submodule — see workflow note below):*
- `ProprietaryFiles/VISSIMserver/DriverModel_RealSim_v2021/` → `DriverModel_RealSim/` (output: `DriverModel_RealSim.dll`)
- `ProprietaryFiles/VISSIMserver/DriverModel_RealSim/` → `DriverModel_RealSim_legacy/` (output: `DriverModel_RealSim_legacy.dll`)
- Update `VISSIMserver.sln` project references
- Update `.vcxproj` output names and the wrapper `.cpp` filenames inside each project

*Migrate VISSIM scenario files (`.inpx` / `.inp0`) currently pointing at the old default to the new legacy name:*
- `tests/CoordMerge/coordMerge.inpx` and `.inp0`
- `ProprietaryFiles/CM11_proj/Data/Vissim/coordMerge.inpx` and `.inp0`

Decision needed during implementation: do we migrate these scenarios to the new int-API default DLL (preferred, since the project supports VISSIM 2022), or just rename the legacy reference? Default plan is **migrate to the new default DLL** so the legacy DLL has zero in-repo consumers.

*Update build infrastructure:*
- `scripts/dispatch/3_vissim_components.bat`
- `scripts/dispatch/7_build_info.ps1`
[...]

Issue #129, opened 2026-03-13
[MAINTENANCE] Consolidate duplicate VISSIM driver model code into shared base


yunlishao-vibe:
## Task Summary

`DriverModel_RealSim.cpp` and `DriverModel_RealSim_v2021.cpp` are nearly identical (~1887 and ~1922 lines respectively) with only minor API-level differences. Extract the shared logic into a common base to eliminate code duplication.

## Why is this needed?

Currently any bug fix or feature must be applied to both files independently. The two files differ in only ~50 lines out of ~1900 (less than 3%), primarily:
- `long` vs `int` type for VISSIM API parameters (v2021 API uses `int`)
- A few global flags added to v2021: `SUB_EGO_ONLY`, `ENABLE_WARMUP`, `isInitialTimeFinished`
- Variable name `simTime` vs `simTime_g`
- Default value of `NEED_SPEED_LIMIT` (true vs false)

The remaining ~97% of code (message handling, socket communication, data processing, FIXS integration) is identical. Multiple `// TODO: add code` stubs appear in the same locations in both files.

### 1. Analyze and document all differences
- Run a full diff and categorize every difference
- Document API-level differences (VISSIM 2022+ vs 2021 driver model API)

### 2. Extract shared logic
Option A: Single shared header `DriverModel_FIXS_Common.h` with inline implementation
Option B: Shared `DriverModel_FIXS_Common.cpp` compiled into both projects via shared source

### 3. Create thin version-specific wrappers
- `DriverModel_RealSim.cpp` → thin wrapper using modern VISSIM API (`long` parameters)
- `DriverModel_RealSim_v2021.cpp` → thin wrapper using legacy VISSIM API (`int` parameters)
- Both include/link to shared common implementation

### 4. Resolve TODO stubs
- The `// TODO: add code` stubs (at least 8 locations in each file) represent unimplemented message field handlers
- Document what each TODO means and either implement or remove with a proper issue reference

### 5. Add Python auto-test
- Add `tests/Python/test_vissim_driver_parity.py` that:
  - Parses both `.cpp` files
  - Verifies that non-API-signature sections are identical (or differ only in known documented ways)
  - Alerts if one file has a change that wasn't applied to the other

## Files Involved
- `ProprietaryFiles/VISSIMserver/DriverModel_RealSim/DriverModel_RealSim.cpp`
[...]

Issue #96, opened 2025-10-31
[MAINTENANCE] Create simple VISSIM testing case


yunlishao:
**Task Summary**  
Develop a minimal test scenario for validating VISSIM integration.

**Why is this needed?**  
Currently lacking simple testing cases makes it difficult to validate VISSIM functionality and conduct quick integration tests.

**Scope of Work**  
- Create a minimal VISSIM test scenario with simple network geometry
- Add to [tests](tests) directory following existing test structure (similar to tests/CoordMerge)
- Include config.yaml, batch file for execution, and VISSIM network file
- Document test execution and expected behavior

**Environment (if version-specific)**  
- VISSIM version: 2022

**Risks or Dependencies**  
None

**Additional Context (optional)**  
- VISSIM-specific documentation: [doc/VISSIMdoc.md](doc/VISSIMdoc.md)
- Reference existing tests in [tests](tests) directory for structure

PR #151, opened 2026-06-06
#96/SpeedLimitLite: bless int-API + legacy CSVs for both VISSIM versions

yunlishao-vibe:
## Summary

Follow-up to #149 and #150. Per the Agent #129/#147 cross-validation findings, the SpeedLimitLite blessed CSVs were inadvertently locked to the long-API DLL path. This PR re-blesses against **both** DLL ABIs so the test is robust to #147's upcoming DLL rename and to PTV's API switch.

Four blessed CSVs are now committed:

| File | VISSIM | DLL ABI | Trace shape |
|---|---|---|---|
| `speed_limit_lite_orig_v2022.csv` | 2022.00-13 | int (`DriverModel_RealSim_v2021.dll`) | 1086 rows, t end=120s |
| `speed_limit_lite_orig_v2022_legacy.csv` | 2022.00-13 | long (`DriverModel_RealSim.dll`) | 380 rows, t end=49.6s |
| `speed_limit_lite_orig_v2026.csv` | 2026 | int | 362 rows, t end=47.7s |
| `speed_limit_lite_orig_v2026_legacy.csv` | 2026 | long | 380 rows, t end=49.6s |

## Related Issues / Tasks

Continues the #96 / milestone 0.9.0 work. PRs #149 and #150 landed the test scaffolding and an initial bless; this PR completes the four-baseline matrix in preparation for #147.

## Type of Change

- [x] New feature (--driver-dll and --api flags)

- `tests/Vissim/SpeedLimitLite/start_vissim.py` — `--driver-dll {int,legacy,<path>}` + `FIXS_DRIVER_DLL` env var, default `int`; pre-flight checks the resolved DLL exists
- `tests/Vissim/SpeedLimitLite/compare_traces.py` — `--api {int,legacy}` flag, default `int`, appends `_legacy` suffix to the picked reference CSV when set to legacy
- `tests/Vissim/SpeedLimitLite/speed_limit_lite_orig_v{2022,2026}.csv` — re-blessed against the int-API DLL (the post-#147 default)
- `tests/Vissim/SpeedLimitLite/speed_limit_lite_orig_v{2022,2026}_legacy.csv` — preserves the previous long-API blessing for the legacy code path
- `tests/Vissim/SpeedLimitLite/README.md` — documents the 2×2 matrix and DLL selection

## Test Cases

```bat
:: int-API on VISSIM 2022 (defaults)
python start_vissim.py
python compare_traces.py

:: long-API on VISSIM 2022
python start_vissim.py --driver-dll legacy
python compare_traces.py --api legacy

:: int-API on VISSIM 2026
python start_vissim.py --progid VISSIM.Vissim.2600 --driver-dll int
python compare_traces.py --vissim-version 2026
```

All four `(version, api)` cells verified self-consistent (0.0000 diff on all 7 fields) in a single clean session.

## Environment

- Python: 3.10 (conda env `realsim_dev`; `pywin32`, `pandas`)
- VISSIM: 2022.00-13 verified; 2026 verified
- DLL build: rebuilt fresh via `scripts/dispatch/3_vissim_components.bat` immediately before the bless

## Checklist

- [x] Code compiles/runs — both bless rounds (int and legacy) completed cleanly on both VISSIM versions
- [x] Tests pass locally — self-compare returns OK across all 7 fields for each of the four cells
- [x] Documentation updated — README rewrites the comparison section with the 2×2 matrix
- [x] Relevant issues linked
- [x] Version-specific changes noted — `--driver-dll`/`--api` map to the two DLL filenames that exist today; after #147 lands the names flip and the test still works because of the abstraction

## Additional Notes

- The long-API DLL (`DriverModel_RealSim.dll` today, `DriverModel_RealSim_legacy.dll` after #147) is compiled from `DriverModel_RealSim.cpp` with long-typed callback signatures. It targets VISSIM ≤ 2020 but still loads in VISSIM 2021+ because x64 LLP64 keeps long and int both 32-bit. It doesn't exercise the SUB_EGO_ONLY / ENABLE_WARMUP / ENABLE_LOG blocks the int-API path added — that's why the legacy traces are shorter (380 rows ending at t=49.6s) than the int-API ones.
- The asymmetry between int-API on 2022 (1086 rows, ego stays for 120s) vs 2026 (362 rows, ego exits at 47.7s) is a real difference in how the two VISSIM versions schedule ego removal under int-API — same `RandSeed=42` but different internal behavior. Exactly the kind of regression the four-baseline split exists to lock down.
- This PR does NOT yet land the COM dispatch fix (apartment-threading + `CLSCTX_LOCAL_SERVER`) recommended in the Agent's findings; that's a separate scope and would be a separate small PR.


yunlishao (approved):
LGTM — all four (vissim_version, api) cells verified self-consistent (0.0000 diff). int-API + legacy split makes the test forward-compatible with #147.

PR #161, opened 2026-06-06
#158/Stage B — XIL ego coupling via FIXS socket protocol

yunlishao-vibe:
## Summary

Stage B of #158 — XIL ego coupling end-to-end via FIXS socket protocol. Builds on Stage A (#160).

**Empirical PASS**: probe at [tests/Vissim/Probes/TrafficLayer_DSProxy_xil/](tests/Vissim/Probes/TrafficLayer_DSProxy_xil/) runs a 3-process orchestration (TrafficLayer in DSProxy mode + fake-CarMaker Python client + VISSIM 2022):

```
=== DSProxy mode (Stage B, issue #158) ===
...
App socket:         server on port 2444 (egoId=ego)
VISSIM_Connect OK
client connected on port 2444
Running 300 ticks (end_time=30.0s at 10 Hz)
tick     0: vehicles=  6 signals= 20 egos=1
ego registered: VISSIM VehicleID=7
tick    25: vehicles= 19 signals= 20 egos=1
...
tick   275: vehicles=124 signals= 20 egos=1
sending shutdown signal to client ...
=== DSProxy mode done (rc=0) ===
```

```
[fake_carmaker] tick    0 sim_time=  0.0 recv_vehicles=  6 ego_x_sent=397.88
[fake_carmaker] tick  275 sim_time= 27.5 recv_vehicles=124 ego_x_sent=672.88
[fake_carmaker] received shutdown signal at tick 300
total ticks: 300, peak received vehicles: 139, ego ran 397.88 -> 696.88 m
```

## Pipeline added

```
fake_carmaker.py  --[VehFullData_t @ port 2444]-->  TrafficLayer.exe
                                                        |
                                                        v
                                                     VISSIM 2022
                                                        |
                                                        v
TrafficLayer  --[VehFullData_t + TrafficLightData_t]-->  fake_carmaker.py
```

Per tick:
1. Push DS egos via `VISSIM_SetDriverVehicles`
2. `VISSIM_GetTrafficVehicles` + `VISSIM_GetSignalStates`
3. Translate `VISSIM_Veh_Data` → `VehFullData_t`, `VISSIM_Sig_Data` → `TrafficLightData_t`
4. `SocketHelper::sendData` to client
5. `SocketHelper::recvData` from client (one VehFullData_t expected)
6. If received id matches `EgoId`, translate to `Simulator_Veh_Data` for next tick — with `Create=true` first frame, `Create=false` after VISSIM round-trips a `VehicleID`

## Related Issues / Tasks

[...]

PR #160, opened 2026-06-06
#158/Stage A — TrafficLayer ↔ DSProxy plumbing

yunlishao-vibe:
## Summary

Stage A of the staged plan in #158. TrafficLayer can now drive a fresh VISSIM 2022 instance via PTV's `DrivingSimulatorProxy.dll` end-to-end when `VissimDSProxySetup.Enable: true` in config.

**Scope deliberately narrow**: no CarMaker, no DriverModel changes, no FIXS-side socket publishing. Stages B–D of #158 build on top of this plumbing.

**Empirical confirmation**: probe scenario at [tests/Vissim/Probes/TrafficLayer_DSProxy/](tests/Vissim/Probes/TrafficLayer_DSProxy/) runs 300 ticks at 10 Hz against PTV's shipped DS example `.inpx`. Vehicles grow 6 → 264 (VISSIM-internal background traffic), signals constant at 20/frame, clean `VISSIM_Disconnect`, exit 0.

```
=== DSProxy mode (Stage A, issue #158) ===
VissimVersion:      2022
SimulatorFrequency: 10 Hz
VISSIM_Connect OK
Running 300 ticks (end_time=30.0s at 10 Hz)
tick     0: vehicles=  6 signals= 20
tick    25: vehicles= 28 signals= 20
tick   275: vehicles=264 signals= 20
calling VISSIM_Disconnect ...
=== DSProxy mode done (rc=0) ===
```

## Related Issues / Tasks

- Stage A of **#158**. Closes the first checkbox.
- No ProprietaryFiles changes in this PR — Stage C's `socketSetup` gating is the dual-PR slice.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Maintenance / Refactor
- [ ] Documentation
- [x] Test case / scenario update (new probe)

## Affected Modules / Components

- [`CommonLib/VissimDSProxyHelper.{h,cpp}`](CommonLib/VissimDSProxyHelper.h) — new. Dynamic-load wrapper around `DrivingSimulatorProxy.dll`. Struct/enum surface mirrors PTV's 2022 header (the 32-line 2022/2026 header diff is doc-only per #156 evidence in PR #159). `connect()` has both `std::wstring` and `std::string` overloads so callers don't have to widen themselves. `defaultDllPathForVissim()` / `connectVersionNo()` exposed at namespace scope for reuse by future probes / stages.
[...]

Issue #158, opened 2026-06-06
[FEATURE] Adopt VISSIM DrivingSimulatorProxy.dll in TrafficLayer (Stages A-D)


yunlishao:
**Is your feature request related to a problem?**
Milestone 0.9.0 needs CarMaker-VISSIM synchronization. Today FIXS hooks VISSIM exclusively through the per-vehicle `DriverModel` DLL path, which is the wrong API for clock-locked external dynamics from CarMaker. The design investigation in #156 selected PTV's `DrivingSimulatorProxy.dll` as the right coupling, with a hybrid (B′) shape that preserves the existing DriverModel for behavior-modifier scenarios (Python CAV controllers).

This issue is the implementation umbrella for that adoption. It replaces #101's original VirtualEnvironment-centric framing (which #157 partially superseded already) and follows from the design half of #156.

**Describe the solution you'd like**
Land the staged plan from [doc/156_drivingsim_dll_design_proposal.md](../blob/maintenance/156_drivingsim_dll_design/doc/156_drivingsim_dll_design_proposal.md) as four sequential PRs.

Decision summary (full rationale in the design doc):
- **Vehicle-state path: B′** — DSProxy in TrafficLayer for ego + readback; DriverModel demoted to behavior-modifier on flagged types
- **Signal source:** `VISSIM_GetSignalStates` from DSProxy (same tick as vehicle data)
- **DriverModel scope under B′:** behavior-cmd-down-only; state-up reporting retired
- **TrafficLayer ↔ VISSIM coupling:** in-process linkage via DSProxy DLL

Staged checklist (each is its own PR; each PR title references this issue and the stage letter):

- [ ] **Stage A — TrafficLayer ↔ DSProxy plumbing.** C++ port of the Python DSProxy wrapper into `CommonLib/VissimDSProxyHelper.{h,cpp}`; `VissimDSProxy:` ConfigHelper schema; TrafficLayer code path that drives VISSIM via DSProxy when enabled. New probe `tests/Vissim/Probes/TrafficLayer_DSProxy/`. No CarMaker, no DriverModel changes. (~1 wk)
- [ ] **Stage B — XIL ego coupling (CarMaker-VISSIM).** Wire CarMaker dyno poses into the DSProxy ego inject path. Refurbish `tests/Vissim/Ipg/` to runnable on CM 13.1.2 with a Parquet reference trace. Fill the VISSIM section of `doc/CarMakerDoc.md`. (~1–2 wk)
[...]

Issue #99, opened 2025-10-31
[MAINTENANCE] Explore FIXS TrafficLayer DLL/shared memory for VISSIM performance


yunlishao:
**Task Summary**  
Explore FIXS TrafficLayer DLL or shared memory approach to resolve VISSIM driver model delayed response issue.

**Why is this needed?**  
Current VISSIM driver model implementation has delayed response issues that impact co-simulation performance. Moving to a DLL-based TrafficLayer or shared memory communication could reduce latency.

**Scope of Work**  
- Investigate root cause of delayed response in current socket-based communication
- Explore TrafficLayer as DLL alternative (in-process communication with VISSIM)
- Investigate shared memory IPC for lower latency
- Prototype and benchmark different approaches
- Compare performance vs current TCP socket implementation
- Document findings and recommend architecture changes

**Environment (if version-specific)**  
- VISSIM version: 2022
- C++ compiler/version: Visual Studio 2019

**Risks or Dependencies**  
- May require architectural changes to TrafficLayer
- DLL approach might need TrafficLayer to be compiled as both executable and library
- Shared memory approach requires synchronization mechanism design

**Additional Context (optional)**  
- Related to performance issue #88: FIXS running slower than real time in co-simulation
- Current architecture uses TCP sockets with TrafficLayer as central hub
- VISSIM integration via driver model DLL: [ProprietaryFiles/VISSIMserver](ProprietaryFiles/VISSIMserver)
- TrafficLayer entry point: [TrafficLayer/TrafficLayer/mainTrafficLayer.cpp](TrafficLayer/TrafficLayer/mainTrafficLayer.cpp)

PR #162, opened 2026-06-06
#158/Stage C — DriverModel + DSProxy coexistence (submodule bump for PF#6)

yunlishao-vibe:
## Summary

Stage C of #158 — DriverModel + DSProxy coexistence. Bumps the `ProprietaryFiles` submodule to pick up the gating fix in `ProprietaryFiles#6`, which moves the `Sock_c.socketSetup` block inside the existing `if (ENABLE_REALSIM)` guard in `DriverModel_FIXS_Common.h`.

## Companion PR

- `ProprietaryFiles#6` (`fix(VISSIMserver): gate socketSetup on ENABLE_REALSIM`) — must merge before this FIXS PR's submodule pointer becomes valid against `main` of the submodule. https://github.com/ORNL-Real-Sim/ProprietaryFiles/pull/6

## Why this unlocks Stage C

The 4-mode coexistence matrix from `tests/Vissim/Probes/DSProxy_DriverModel_coexist/` in PR #159 isolated the root cause:

| Mode | `VISSIM_Connect` |
| --- | --- |
| no DriverModel | True (13.3 s) |
| PTV stock DriverModel sample | **True (13.5 s)** — proved DSProxy + DriverModel CAN coexist |
| missing DLL path | False (14.0 s) |
| FIXS `DriverModel_RealSim.dll` (pre-patch) | False (20.0 s) — root cause: unconditional `socketSetup` blocked DSProxy's handshake watchdog |

The `ProprietaryFiles` patch makes the failing row pass: with the gate in place, `socketSetup` no longer runs when `EnableRealSim: false` in the par-file. The DriverModel still loads, still receives every per-vehicle callback, still returns `USE_INTERNAL_MODEL=1` (so VISSIM continues its Wiedemann integration), but no socket is bound.

The empirical chain is therefore:

> Stage 2 probe (PR #159) showed PTV stock DriverModel + DSProxy coexists →
> PF#6 reduces FIXS DriverModel to the same shape (no socket) when `EnableRealSim: false` →
> Stage C is unlocked: a single .inpx can have FIXS DriverModel-hooked vehicle types AND a DSProxy-driven ego.

## Related Issues / Tasks

- Stage C of **#158**. Checks the third checkbox.
- Stacks on **#160** (Stage A) and **#161** (Stage B). When those merge to `dev_v0.9.0`, this branch needs a rebase but the Stage C diff itself is independent (just a submodule pointer bump).
- Companion: `ProprietaryFiles#6`.

## Type of Change

- [x] Bug fix (per-VISSIM-instance coexistence between FIXS DriverModel and PTV DSProxy was broken — root cause + remedy in the PF PR)
- [ ] New feature
- [x] Maintenance / Refactor (FIXS-side: submodule bump only)
- [ ] Documentation

- `ProprietaryFiles/` — submodule pointer bumped to the PF#6 branch tip. **No FIXS-side C++ changes in this commit.**

The TrafficLayer-side "simplified handshake for behavior_only mode" called out in the design doc is **not needed** in practice: with `EnableRealSim: false` the DriverModel doesn't open a socket at all, so TrafficLayer's existing handshake logic is unaffected. The conceptual `DriverModelMode: behavior_only` flag from the design proposal is achieved via the existing `EnableRealSim: false` semantics — no new YAML knob required. The design doc was over-prescriptive on this; the empirical fix is simpler than the proposal.

## Test Cases

Pre-merge verification depends on `ProprietaryFiles#6` landing first. Once both submodule pointer and PF#6 are in place:

1. Build: `scripts\dispatch\3_vissim_components.bat` (Release x64) — confirmed working with the patched submodule on this branch
2. Coexistence smoke test: re-run the 4-mode matrix in `tests/Vissim/Probes/DSProxy_DriverModel_coexist/` (from PR #159) with a new 5th mode `real_drivermodel_behavior_only` (par-file with `EnableRealSim: false`) and confirm `VISSIM_Connect` succeeds in ~14 s — same as `ptv_stock`. This 5th-mode test is deferred to a follow-up because PR #159's probe hasn't merged to `dev_v0.9.0` yet; testing it now would require cross-branch dependency management.

The existing scenarios that set `EnableRealSim: true` are **bit-identical** in behavior — the patch only changes the path when the existing guard already routed differently. No regression risk.

## Environment

- Python version: n/a (no Python touched in this PR)
- VISSIM version: 2022 (primary; 2026 untested but the patch is VISSIM-version-agnostic)
- IPG CarMaker version: n/a (no CarMaker code touched)
- Windows: 11 24H2
- Compiler: VS 2022, v143 toolset, Release x64

## Checklist

- [x] Code compiles/runs as expected (DriverModel DLLs build clean against the patched ProprietaryFiles)
- [ ] Tests pass locally — deferred: 5th-mode probe needs PR #159 merged first
- [x] Documentation is updated (the rationale is in PF#6's inline comment block + this PR body)
- [x] Relevant issues are linked (Stage C of #158; ProprietaryFiles#6; references #159 for evidence, #160 + #161 for stack)
- [x] Version-specific changes are noted (VISSIM 2022 primary; not version-specific)

## Additional Notes

**Why no `DriverModelMode: behavior_only` flag**: the design doc in PR #159 proposed adding a config flag with that name. Implementing PF#6 made the flag unnecessary — `EnableRealSim: false` already achieves the exact behavior the flag was supposed to provide (DriverModel loads, processes callbacks, returns `USE_INTERNAL_MODEL=1`, opens no socket). Adding a redundant knob would just bloat the config schema. If reviewers think the explicit name is worth the cost (clearer config intent), the flag can land as a doc-only alias in a follow-up — but the working semantics are already there.

**Stage D status**: Stage D (signal source migration) is **architecturally absorbed by Stage A**. The DSProxy code path already sources signals from `VISSIM_GetSignalStates` (verified in the Stage A probe). The non-DSProxy code path retains its existing COM-based enumeration because there's no DSProxy DLL loaded to migrate from. No code change needed — will note this in #158 when reporting Stage D as done.

### DIFF
--- a/CommonLib/ConfigHelper.cpp
+++ b/CommonLib/ConfigHelper.cpp
@@ -299,11 +299,17 @@ int ConfigHelper::getConfig(string configName) {
 	}
 
 	// figure out TrafficLayer IP
-	if (XilSetup.EnableXil) {
+	// Both branches assume the chosen subscription has at least one entry,
+	// which is true for every shipped scenario today. Guarding the empty
+	// case here lets a Stage A DSProxy probe config — which deliberately
+	// has no subscriptions yet — load without an access violation. The IP/
+	// port stay at default-constructed values; consumers further down the
+	// main path require at least one subscription anyway.
+	if (XilSetup.EnableXil && !XilSetup.VehicleSubscription.empty()) {
 		SimulationSetup.TrafficLayerIP = get<2>(XilSetup.VehicleSubscription[0])[0];
 		SimulationSetup.TrafficLayerPort = get<3>(XilSetup.VehicleSubscription[0])[0];
 	}
-	else {
+	else if (!XilSetup.EnableXil && !ApplicationSetup.VehicleSubscription.empty()) {
 		SimulationSetup.TrafficLayerIP = get<2>(ApplicationSetup.VehicleSubscription[0])[0];
 		SimulationSetup.TrafficLayerPort = get<3>(ApplicationSetup.VehicleSubscription[0])[0];
 	}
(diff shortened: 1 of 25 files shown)


yunlishao-vibe:
## Update: integrated regression probe added

Added [`tests/Vissim/Probes/TrafficLayer_DSProxy_coexist_xil/`](tests/Vissim/Probes/TrafficLayer_DSProxy_coexist_xil/) on this branch as an empirical close-out for the claim that PF#6 unlocks DSProxy + FIXS DriverModel coexistence. The previously-deferred "5th mode" empirical test now exists and passes.

**Pipeline tested**:
```
python_ego.py (egoctrl-style trajectory)
    -> TrafficLayer (DSProxy mode) via FIXS binary protocol on port 2444
    -> VISSIM 2022 with Car (type 100) hooked to FIXS DriverModel_RealSim.dll (EnableRealSim:false)
```

**All 7 invariants PASS**:

| # | Invariant | Result | Stage covered |
| --- | --- | --- | --- |
| 1 | `B_ego_visible_near_pushed_pose` | 60/60 on-link frames | B |
| 2 | `B_ego_registered_per_tl_log` | TL stdout has "ego registered: VISSIM VehicleID=..." | B |
| 3 | `B_background_traffic_grew` | vehicles 45 → 76 | B |
| 4 | `B_tls_received` | 85 ticks completed | B |
| 5 | `C_drivermodel_flagged_type_present` | 85/85 ticks had Car in readback | **C** |
| 6 | `C_drivermodel_init_evidence` | `DriverModelError.txt` size 319 B (multi-entry) | **C** |
| 7 | `C_no_traffic_layer_socket_error` | no "Error: initialize connection to Traffic Layer" | **C** |

This is the first probe where the **actual FIXS-built `DriverModel_RealSim.dll`** (not just PTV stock) coexists with DSProxy through TrafficLayer. PF#6 (Real-Sim-XIL/ProprietaryFiles#6) is empirically validated end-to-end.

Note: this same probe doubles as the stronger Stage B regression test (the existing `TrafficLayer_DSProxy_xil/` probe only ran a constant-velocity ego with no invariant checks). The egoctrl-style trajectory + invariant matrix here is the regression contract Stage B was missing.

`out/summary.json` is checked in as a regression baseline.

yunlishao-vibe:
Superseded. The #173 headless investigation showed (a) the real VISSIM_Connect blocker is the DriverModel DRIVER_DATA_ALLOW_MULTITHREADING declaration, not the socketSetup gating this PR diagnoses, and (b) this verifies only the INERT co-load (EnableRealSim:false, no socket). The active Stage C relay is reworked under #99. Closing.

PR #163, opened 2026-06-06
#158/Stage B+ — CAV controller relay end-to-end (DM speed override works)

yunlishao-vibe:
## Summary

Stage B+ of issue #158 (B' doctrine, CAV controller path): a Python client
drives both an ego pose into DSProxy and per-vehicle behavior commands
into the FIXS DriverModel over a single TrafficLayer socket. End-to-end
pipeline now works through to Wiedemann; the CAV speed override is
empirically applied.

Companion PR (ProprietaryFiles): Real-Sim-XIL/ProprietaryFiles#7

## Type of change

- [x] New feature
- [ ] Bug fix
- [x] Test / probe addition

## What's wired

```
python_ego_cav.py  --[VehFullData @ port 2444]--> TrafficLayer (DSProxy mode + DM relay)
                                                       |
                                       +---------------+---------------+
                                       v                               v
                                  DSProxy.dll                  DriverModel socket (1337)
                                       |                               |
                                       v                               v
                              VISSIM_SetDriverVehicles         DriverModel_FIXS_Common.h
                              (ego pose every tick)            DRIVER_DATA_TIME callback
                                                                       |
                                                                       v
                                                              Wiedemann integration
                                                              w/ overridden v_desired
```

## Root Cause + Fix (Stage B+ tick-0 deadlock)

Two independent bugs were masking each other. Both fixed in this PR.

**Bug 1 — TrafficLayer skipped DriverModel I/O on tick 0.** The earlier
guard `if (relayDM && dmSock > 0 && tick >= 1)` was based on the assumption
that the DM's first-tick path takes an "isVeryFirstStep" branch that does
no I/O. That branch exists, but `DRIVER_DATA_TIME` is in fact called
*multiple times* per VISSIM tick, and the second call does send + recv
even on tick 0. With TL skipping I/O on tick 0, DM blocks waiting for
TL's command, and TL's `getTrafficVehicles` blocks waiting for VISSIM
tick 0 to complete. Resolved by removing the `tick >= 1` guard.

**Bug 2 — `SUB_EGO_ONLY` was hardcoded `true` in the FIXS DriverModel.**
With no subscribed vehicle list (the DSProxy + DM relay case), the
DriverModel skipped per-tick I/O entirely, even after TL did its
first-tick recv. Resolved by adding a `SimulationSetup.SubEgoOnly` knob
to `ConfigHelper` (default `true` = legacy) and reading it into the DM's
global in the PF companion PR.

## What changed

[...]

Issue #117, opened 2026-01-28
[FEATURE] Design execution timing mechanism rehaul


yunlishao:
**Is your feature request related to a problem?**

As Real-Sim FIXS expands to support more XIL configurations and multi-rate co-simulation scenarios, the execution timing architecture needs a deliberate redesign. A formalized timing framework would enable real-time pacing, multi-rate execution, and predictable behavior under load.

Core areas to address:
- **Execution timing contract**: Establish a formal contract defining how simulation steps are scheduled, what timing guarantees are provided, and how components declare their timing requirements.
- **Deadline miss behavior**: Allow users to declare what should happen when a simulation step cannot meet its timing deadline (e.g., skip, catch up, warn, halt).
- **Multi-rate co-simulation**: Support components running at different rates (e.g., traffic at 10 Hz, vehicle dynamics at 100 Hz) with well-defined synchronization points.

**Describe the solution you'd like**

This is a **design exploration** issue. The goal is to investigate the three key ideas above and produce a design/implementation plan.

Scope of design work:
1. Explore what an execution timing contract should look like for FIXS — how should components declare their timing needs via `config.yaml`?
2. Explore user-declared deadline miss policies — what behaviors make sense, and how should they be configured?
3. Explore multi-rate support — how should TrafficLayer orchestrate components running at different frequencies, considering both SUMO (TrafficLayer-driven stepping) and VISSIM (DLL-driven stepping)?
4. Consider impact on existing XIL integrations (CarMaker, Simulink, Carla).
5. Identify sub-tasks to be filed as separate implementation issues under milestone 0.10.0.

**Output:**
- Design document with implementation plan (markdown in `doc/` or wiki)
- List of follow-up implementation issues for milestone 0.10.0

**Describe alternatives you've considered**

- Keeping current execution model and handling timing externally per-component — fragments timing logic and makes system behavior harder to reason about.
- Incremental pacing additions without a broader timing contract — doesn't address multi-rate needs or deadline miss policies.

**Additional context**

Key files relevant to this design:
- `TrafficLayer/TrafficLayer/mainTrafficLayer.cpp` — main simulation loop
- `CommonLib/TrafficHelper.h` — `SIM_STEP`, `runOneStepSimulation()`
- `CommonLib/TrafficHelper.cpp` — step execution, SimulationMode handling
- `CommonLib/ConfigHelper.h` — configuration parsing

[...]

PR #165, opened 2026-06-06
#158/Stage B multi-port subscription routing

yunlishao-vibe:
## Summary

Stacks on Stage C (#162). Refactors `DSProxyMode.cpp` to honor every entry in `ApplicationSetup.VehicleSubscription`, not just `[0]`. Each `(subscription, port)` tuple expands into its own server socket and per-tick filter.

Stage B (#161) was constrained to `VehicleSubscription[0].port[0]` as a simplification — fine for the single-CarMaker case but limiting for scenarios with concurrent consumers (CarMaker dyno + Python observer + Carla viz on different ports).

## How it works

1. `resolveAppSockets(config)` at startup builds a `std::vector<AppSocketConfig>`. One entry per (subscription, port) tuple.
2. SocketHelper opens every port via `socketSetup(vector<int> ports)`. TrafficLayer waits for **all** clients before entering the tick loop.
3. Per tick:
   - Pre-translate `VISSIM_Veh_Data` → `VehFullData_t` once.
   - For each subscribed socket: apply per-port `publishesVehicle` filter, send vehicles + signals.
   - For each subscribed socket: recv `VehFullData_t`. The canonical ego (first non-empty `egoId` across all subscriptions) becomes the next tick's DSProxy ego inject.
4. Shutdown: send `simState=0` to every connected client.

## Per-port filter rule (`publishesVehicle`)

```
if type == "vehicleType" AND non-empty ids → filter by vehicle TYPE
otherwise                                    → publish all
```

**Important**: `type: ego` is intentionally NOT a publish filter. The ego id (`'ego'`) is used on the recv path to identify which incoming `VehFullData_t` is the ego inject for DSProxy. Filtering outbound by id would never match VISSIM's auto-assigned numeric `VehicleID` and would break every existing scenario. This matches `mainTrafficLayer`'s existing `ENABLE_VEH_SIMULATOR == false` behavior — the subscription tuple is for picking `(ip, port)` pairs, not for per-vehicle filtering.

## Regression-safe

Re-ran `tests/Vissim/Probes/TrafficLayer_DSProxy_xil/` (Stage B's single-port `type: ego` subscription probe) against this branch — 300 ticks PASS, peak 289 vehicles, ego registered as VehicleID=9, clean shutdown rc=0. Identical to the Stage B baseline.

## Related Issues / Tasks

- Stacks on **#162** (Stage C). Multi-port routing is a Stage B refinement called out as deferred in #161 / #162 PR bodies.
- Independent of **PR #163** (Stage B+ CAV controller) and **PR #164** (Python TLS parser fix). Can land in any order relative to those.

## Type of Change

- [ ] Bug fix
- [x] New feature (multi-subscription support in DSProxy mode)
- [x] Maintenance / Refactor (single-port path is now the N=1 case of the general path)
- [ ] Documentation

- `TrafficLayer/TrafficLayer/DSProxyMode.cpp` — `AppSocketConfig` is now per-tuple; `resolveAppSockets()` returns a vector; tick loop iterates the vector; per-tuple `publishesVehicle` filter

## Test Cases

- **Regression**: re-ran Stage B probe (`TrafficLayer_DSProxy_xil/`) — PASS, 289 peak vehicles, identical to baseline
- **Build**: TrafficLayer.exe Release x64 builds clean
- **Multi-port end-to-end with concurrent clients** — NOT in this PR. The code path is exercised by the `vector<>`-based loop but a dedicated probe with 2+ Python clients connecting simultaneously is small follow-up work, called out below.

## Environment

- Python: 3.10 / `realsim_dev`
- VISSIM: 2022 (single-port regression verified; multi-port code path is version-agnostic)
- Windows: 11 24H2
- Compiler: VS 2022 v143

## Checklist

- [x] Tests pass locally (Stage B regression — single subscription on port 2444)
- [x] Documentation is updated (inline comments explain filter rule + receive-side ego semantics)

## Additional Notes — carryover

- **Multi-port end-to-end probe**: dedicated test scenario with 2+ subscriptions on different ports + 2+ Python clients. Would validate per-port filtering and concurrent send/recv.
- **Per-port signal/detector filtering**: signals are currently global (every subscribed client sees all 20/tick from VISSIM_GetSignalStates). The existing main `TrafficLayer` does per-port signal subscription filtering; we could mirror that here when a real signal-filter scenario appears.
- **Detector data path**: wired through `msgHelper.DetDataSend_um` for protocol completeness but no filter applied. Fine because Stage A/B/C probes don't exercise detectors.


yunlishao-vibe:
Folded in a multi-port cosim probe: `tests/Vissim/Probes/TrafficLayer_DSProxy_multiport_xil/`.

Two Python clients on two ports, all 6 invariants PASS:

- both clients connect + lockstep ticks
- port 2445 (`type: vehicleType, id: ['100']`) sees ONLY type='100' vehicles
- port 2444 (`type: ego`) sees the full mix of types (100, 200, 300, 400)
- per-tick TLS count identical across both ports

The 6-vehicle delta between ports at tick 0 = the non-Car vehicles on the network = empirical proof that `publishesVehicle()` is actually filtering, not just passing everything through.

Commit: db890362

yunlishao-vibe:
Folded in #117 absorption-readiness annotations:

- Per-tick loop now uses PHASE 1-7 markers matching the canonical FIXS tick flow (`doc/fixs_tick_flow.md`, lands via #161 — this branch picks it up at rebase).
- `publishesVehicle()` annotated as the canonical **PHASE 3 routing-rule** implementation: when #117's orchestrator lands, this function's (subscription, message) -> publish-or-skip decision becomes a row in the orchestrator's routing table; body unchanged.
- Header doc explains how the multi-port loop iterating `appSocks` mirrors mainTrafficLayer's legacy single-list iteration. Same SocketHelper / MsgHelper contract; different iteration target.

No behavior change. All 6 probe invariants still PASS. Commit: 9ce967cc

yunlishao-vibe:
Keeping open as a future refactor. The integrated #168/#172 demo line (now in dev_v0.9.0) uses simpler single-client app-socket routing; this multi-port subscription routing is the proper generalization to adopt later. Not blocking the demos.

yunlishao (dismissed):

yunlishao-vibe:
Rebased onto current `dev_v0.9.0` (415 commits of drift) and re-verified end-to-end.

**Why this wasn't a mechanical merge.** `DSProxyMode.cpp` grew substantially on `dev_v0.9.0` after this branch forked — DriverModel relay + CAV-command split, the Plan A signal socket (#172), the compass/math heading conversion, `vehicleClass` mapping, and the ego-id re-stamp on publish. This branch's June copy predates all of that, so taking either side wholesale would have regressed something. Resolution keeps every `dev_v0.9.0` behavior and generalizes the single app socket into a vector:

- `resolveAppSocket()` → `resolveAppSockets()`: one `AppSocketConfig` per `(subscription, port)` tuple; the legacy single-client path is the N=1 case.
- `publishesVehicle()` per-port outbound filter. Only `type: vehicleType` filters; `type: ego` stays a recv-side identifier (its VISSIM id is numeric and would never match `'ego'`).
- `bindListener` now runs once per app port, still **before** `VISSIM_Connect` — that ordering is load-bearing, since the FIXS DriverModel connects during that handshake. `TrafficSignalPort` is validated against every app port, not just one.
- Per tick: translate VISSIM state to `VehFullData_t` once, then send to each client through its filter, recv from each client, first inbound `id == canonicalEgoId` becomes the DSProxy ego inject. Shutdown sends `simState=0` to every client.

**Two probe fixes were needed before it could run at all.** The committed `config.yaml` used the retired `VissimDSProxySetup: Enable:` key, so TrafficLayer silently fell through to the *legacy* path (it printed `Client ports: 2444, 2445` and never entered DSProxy mode) — now `VissimSetup: EnableDSProxy:`. And `NetworkFile` must be absolute, so `run_multiport_xil.bat` now generates `config.runtime.yaml` resolved to the current checkout, the same convention the CMoffice demos use, instead of hardcoding another worktree's path.

**Verification (VISSIM 2022, this build):**

Multi-port probe — 6/6 invariants PASS:
```
PASS both_clients_connected              (ego=True observer=True)
PASS both_clients_no_error
PASS tick_aligned                        (ego_ticks=80 obs_ticks=80)
PASS observer_filtered_to_Car            (observer saw types: ['100'])
PASS ego_port_unfiltered                 (ego port saw types: ['100', '300', '400'])
PASS signals_count_aligned_across_ports  (80/80 ticks identical; 20 TLS/tick)
```
The per-tick vehicle delta between ports (58 vs 53 at tick 60) is the non-Car traffic — the filter is selecting, not passing everything through. TLS now parses on the Python side (20/tick, was 0 in the original run) since #163 landed.

N=1 regression — single ego subscription driving `fake_carmaker`: 300 ticks, peak 288 vehicles, clean shutdown rc=0. To be sure that was not just "passes", I built a **control** of unmodified `dev_v0.9.0` `DSProxyMode` in the same tree and ran the same probe: identical tick count, identical peak, and identically the ego is created at tick 0 but its `CreateID` never round-trips (`egos=1` then `egos=0` forever). That ego-registration gap is pre-existing `dev_v0.9.0` behavior on the PTV DS-example network, not something this PR introduces.

Filed #292 for the stale probe configs that surfaced during this work.

yunlishao (approved):

Issue #169, opened 2026-06-11
[MAINTENANCE] FP-accumulation timing gates (fabs(simTime*rate - round) < 1e-5) silently freeze over long runs


yunlishao:
**Task Summary**
Several "fire once per Nth tick" timing gates across FIXS decide when to fire by testing whether `simTime * rate` lands on an integer within a fixed absolute tolerance — `fabs(simTime*rate - round(simTime*rate)) < 1e-5` (sometimes written `ceil(x - 0.5)`). This construct silently stops firing on long runs as accumulated floating-point error in `simTime` grows past the fixed tolerance. One instance already caused a hard failure in the #168 CarMaker↔VISSIM co-sim; the rest are the same bug, latent at larger run-times. Convert them all to an exact integer-slot check.

**Why is this needed?**
The test asks *"did `simTime*rate` land on an integer, within 1e-5?"*. `simTime` accumulates FP rounding error that grows with run length; multiplied by `rate`, that error eventually exceeds `1e-5` **permanently**, after which the gate never fires again and whatever it guards freezes — no crash, no log. Time-to-failure scales roughly as 1/rate², so the only instance that bit us in practice is the one with a large, config-driven multiplier.

Observed failure (#168): the traffic refresh/interpolation gate in `VirEnvHelper.cpp` (rate = `1/TrafficRefreshRate` = **1000** at the default 1 ms refresh) crossed the tolerance at **~715 s**, freezing **all** traffic-vehicle interpolation fed to CarMaker — ego and background traffic drifted ~190 m apart by 1000 s while neither side errored. Root-caused by logging the same gated quantity from the writer and reader loops and watching their timelines diverge (715 s vs 1000 s). Fixed on `feature/168_cm_office_xil` in commit `56a87ca7` with the integer-slot pattern below; this ticket tracks the remaining instances plus the dead-but-revivable ones.

### Progress
- [x] `CommonLib/VirEnvHelper.cpp:826` — refresh/interpolation gate — **fixed in `56a87ca7`** (integer-slot `llround`), verified over a full 1000 s headless run (offset flat, was ~190 m)
- [ ] `CommonLib/VirEnvHelper.cpp:402` — `receive_realsim` gate
- [ ] `CommonLib/VirEnvHelper.cpp:713` — `send_ego` gate
[...]

Issue #167, opened 2026-06-08
[FEATURE] Absorb VISSIM DSProxy mode into the XIL orchestrator (remove early-return bypass)


yunlishao:
**Is your feature request related to a problem?**

Issue #158 (VISSIM DSProxy adoption) ships into milestone 0.9.0 using an **early-return bypass** pattern in `mainTrafficLayer.cpp`:

```cpp
if (Config_c.VissimSetup.EnableDSProxy) {
    return FIXS::DSProxy::runDSProxyMode(Config_c);   // exits, never falls through
}
// legacy DriverModel-socket main loop below (NEVER RUNS when DSProxy is on)
```

`runDSProxyMode` has its own parallel per-tick loop that re-implements PHASES 1-7 (see [`doc/fixs_tick_flow.md`](../blob/dev_v0.9.0/doc/fixs_tick_flow.md)) outside the legacy main loop. This was the deliberate scope-protecting choice for 0.9.0: avoid touching the legacy SUMO + DriverModel-socket paths while shipping DSProxy quickly.

Once #117 (XIL orchestrator) lands in milestone 0.10.0, the bypass becomes unnecessary — the unified orchestrator loop can drive both the legacy DriverModel-socket path AND the DSProxy path through the same pub/sub matrix. This issue tracks the absorption.

**Describe the solution you'd like**

Replace the early-return bypass with adapter registration; let the unified main loop drive both paths. Concretely:

1. **Delete `runDSProxyMode`'s parallel tick loop.** The body lifts into the orchestrator's central loop (which already does PHASES 3-6).
2. **Replace the dispatch in `mainTrafficLayer.cpp`**:
   ```cpp
   // BEFORE (bypass — 0.9.0)
   if (Config_c.VissimSetup.EnableDSProxy) {
       return FIXS::DSProxy::runDSProxyMode(Config_c);
   }
   // ... legacy DriverModel-socket setup + main loop ...

   // AFTER (absorbed)
   if (Config_c.VissimSetup.EnableDSProxy) {
       orchestrator.registerTrafficSimAdapter<DSProxyAdapter>(Config_c.VissimSetup);
   } else {
       orchestrator.registerTrafficSimAdapter<DriverModelSocketAdapter>(...);
   }
   return orchestrator.run();
   ```
[...]

Issue #156, opened 2026-06-06
[MAINTENANCE] Investigate VISSIM DrivingSimulator.dll and design CarMaker-VISSIM sync workflow


yunlishao:
**Task Summary**
Explore VISSIM's `DrivingSimulator.dll` interface (the API PTV exposes for tightly-coupled external simulators / driving simulators) and design how FIXS should use it to support CarMaker–VISSIM synchronization for Milestone 0.9.0. Today FIXS hooks VISSIM exclusively through the **DriverModel DLL** path (`DriverModel_RealSim.dll` / `DriverModel_RealSim_v2021.dll`, see [ProprietaryFiles/VISSIMserver](ProprietaryFiles/VISSIMserver) and [scripts/dispatch/3_vissim_components.bat](scripts/dispatch/3_vissim_components.bat)). It is unclear whether CarMaker–VISSIM sync should:

1. **Co-exist** — keep DriverModel for per-vehicle traffic control and add DrivingSimulator only for the CarMaker ego coupling, or
2. **Migrate** — move all FIXS↔VISSIM interaction onto DrivingSimulator.dll, or
3. **Hybrid** — wrap both behind a single FIXS-side interface in `TrafficHelper`.

This ticket is the investigation + design ticket; an implementation ticket will follow once the workflow is chosen.

**Why is this needed?**
- Milestone 0.9.0 includes CarMaker–VISSIM sync (parent feature: #101).
- VISSIM's DriverModel API is a *per-vehicle driver-behavior* hook; it is not the API PTV documents for clock-locked external dynamics from tools like CarMaker. The DrivingSimulator.dll path is generally the supported route for that use case, but FIXS has never used it.
- We need to know what DrivingSimulator.dll actually exposes (entry points, time-step contract, vehicle injection model, signal state, coordinate frames) before we can scope the implementation or commit to a wire format that survives `VehDataMsgDefs.h` / [CommonLib/MsgHelper.h](CommonLib/MsgHelper.h).
- Picking the wrong DLL path early would mean redoing the CarMaker side and likely re-touching [CommonLib/TrafficHelper.h](CommonLib/TrafficHelper.h) — which is already on the debt list (~1743 lines, milestone 0.8.0 refactor target).

**Scope of Work**

*Investigation deliverables (no FIXS code changes required):*
[...]

Issue #101, opened 2025-10-31
[FEATURE] Modify VirtualEnvironment to support CarMaker-VISSIM synchronization


yunlishao:
**Feature Description**  
Extend VirtualEnvironment component to support synchronization between CarMaker and VISSIM co-simulation scenarios.

**Why is this needed?**  
Currently VirtualEnvironment focuses on Carla integration. Adding CarMaker-VISSIM sync support will enable:
- Synchronized visualization between CarMaker ego vehicle and VISSIM traffic
- Coordinated scenario playback across both simulators
- Enhanced XIL testing with VISSIM background traffic and CarMaker vehicle dynamics

**Scope of Work**  
- Extend [VirtualEnvironment/VirtualEnvironment](VirtualEnvironment/VirtualEnvironment) to handle CarMaker-VISSIM sync
- Add configuration options in config.yaml VirtualEnvironmentSetup section
- Implement synchronization logic for:
  - Time synchronization between CarMaker and VISSIM
  - Vehicle state exchange and coordinate transformation
  - Scenario event coordination
- Update VirtualEnvironment message handling for CarMaker data
- Create test scenario demonstrating CarMaker-VISSIM co-simulation
- Update documentation in [doc/CarMakerDoc.md](doc/CarMakerDoc.md) and [doc/VISSIMdoc.md](doc/VISSIMdoc.md)

**Environment (if version-specific)**  
- VISSIM version: 2022
- CarMaker version: 13.1.2
- C++ compiler/version: Visual Studio 2019

**Risks or Dependencies**  
- Requires understanding of CarMaker HIL interface
- May need coordinate system transformation utilities
- Should align with VISSIM driver model overhaul architecture (#99)
- Depends on stable TrafficLayer-VISSIM communication

**Additional Context (optional)**  
- Current VirtualEnvironment implementation: [VirtualEnvironment/VirtualEnvironment](VirtualEnvironment/VirtualEnvironment)
- CarMaker integration docs: [doc/CarMakerDoc.md](doc/CarMakerDoc.md)
- ConfigHelper handles VirtualEnvironmentSetup: [CommonLib/ConfigHelper.h](CommonLib/ConfigHelper.h)

PR #170, opened 2026-06-12
#168/Real CarMaker office co-sim validation of Stage B DSProxy pipeline

### LINKED ISSUE #168: [FEATURE] Real CarMaker office validation of Stage B DSProxy pipeline (replace fake_carmaker.py)

#158 Stage B (PR #161) wires up the FIXS XIL ego-coupling pipeline (TrafficLayer drives VISSIM via DSProxy; an external client sends ego pose, receives traffic state), but validates it only with `fake_carmaker.py` — a Python stand-in. The actual `tests/Vissim/Ipg/` refurbishment for real CarMaker was explicitly deferred from #161's PR scope.

This ticket tracks the **minimum-viable real-CarMaker validation** of the same Stage B pipeline: replace fake_carmaker with the real CarMaker office GUI driving a simple TestRun, prove the round-trip works end-to-end, no Simulink, no dSPACE.

Replace `fake_carmaker.py` with real CarMaker in a focused new probe under `tests/Vissim/Probes/`. Concretely:

1. **Pick a simple VISSIM network.** Either reuse PTV's shipped `driving_simulator_test.inpx` (already used by `tests/Vissim/Probes/TrafficLayer_DSProxy_xil/`) or a custom minimal `.inpx` (one straight link, maybe one signal). Whatever's simplest to verify visually.
2. **Build a minimal CarMaker office TestRun** that:
   - Loads a flat road or matching VISSIM network geometry
   - Has one ego vehicle with default dynamics (no Simulink controller, no CM4SL)
   - Drives the ego via a basic IPGDriver trajectory (steady straight-line, or a curve)
3. **Wire CarMaker to TrafficLayer over the FIXS socket protocol** the same way `fake_carmaker.py` does — same port, same `VehFullData_t` exchange, same per-tick handshake. The CarMaker-side bridge is the User.c hook (existing FIXS pattern from `tests/Vissim/Ipg/`).
4. **Land a probe directory** `tests/Vissim/Probes/TrafficLayer_DSProxy_CMoffice/` with: README, the CarMaker project, runner batch file, and a simple validation script that confirms the round-trip.

**Acceptance criteria**

- [ ] CarMaker office launches and connects to TrafficLayer on the configured FIXS port.
- [ ] VISSIM (driven by TrafficLayer via DSProxy) receives the ego pose from CarMaker every tick — ego is visibly present at the CarMaker-driven position.
- [ ] CarMaker receives VISSIM traffic state every tick (vehicles and signals visible in the CarMaker scene, or at least logged in a sink).
- [ ] Run completes for ≥30 seconds without socket drops, hangs, or zombie processes.
- [ ] README documents the setup steps end-to-end (CarMaker install location, VISSIM install, TestRun load, run order).
- [ ] Cross-reference from `doc/CarMakerDoc.md` to the probe.

- **Stay on fake_carmaker.py only** — rejected; we need empirical proof that the FIXS protocol works against the actual CarMaker User.c hook, not just a Python equivalent. Wire format bugs can hide there.
- **Validate with Simulink + dSPACE first** — rejected for scope; that's the full real-time XIL stack and adds many failure modes. Office-only first.
- **Refurbish the existing `tests/Vissim/Ipg/` scenario** — possible, but that's a CoordMerge-style multi-vehicle test from the SUMO era. A fresh minimal probe is cheaper than fixing the legacy scenario for the first proof-of-concept.

- VISSIM version: 2022 (primary)
- IPG CarMaker version: 13.1.2 (CarMaker office; **no Simulink**, **no CM4SL**, **no dSPACE**)
- Windows: 11 24H2
- C++ compiler/version: Visual Studio 2022

**Additional context**

### Out of scope (deferred)

- **Simulink controller** — keep CarMaker on its built-in IPGDriver. No `.slx`, no MATLAB.
- **CM4SL / Simulink Coder** — explicitly not used.
- **dSPACE real-time** — office-only validation; real-time XIL is a follow-up.
- **`tests/Vissim/Ipg/` legacy scenario refurbishment** — separate ticket if needed; this probe is fresh.
- **Multi-ego scenarios** — single ego only.
- **Performance / real-time pacing** — wall-clock pacing is fine; deadline-bounded XIL is #117's job.

### Why this matters now

- Validates the FIXS DriverModel-on-CarMaker User.c hook against real CarMaker before Stage C (#162) builds on top with the DriverModel coexist mode.
- Closes the unspoken gap in #161 where Stage B "supports CarMaker" but nobody has empirically tested it.
- Produces a known-good reference scenario that future XIL work (real-time, multi-ego, Simulink controllers, dSPACE) can incrementally build on without re-proving the basic pipeline.

### Cross-references

- **#158** — VISSIM DSProxy adoption umbrella; this is the deferred CarMaker validation piece of Stage B.
- **#161** — Stage B PR (the Python fake_carmaker probe that this ticket promotes to real CarMaker).
- **`tests/Vissim/Probes/TrafficLayer_DSProxy_xil/`** — the existing fake_carmaker probe; this new probe parallels its shape.
- **`doc/CarMakerDoc.md`** — the VISSIM section of this doc gets a real-CarMaker example added.
- **#101 / #156** — original CarMaker-VISSIM sync intent; partially fulfilled by closing this.
- **#117** — future XIL orchestrator (the real-time deadline + replay concerns this ticket explicitly defers).
- **#167** — DSProxy absorption into orchestrator; this ticket's probe should be one of the regression scenarios that absorption preserves.

### Notes

- This is **NOT a research-mode issue**. The pipeline already exists (Stage B PR validates it with Python); this ticket is bringing real CarMaker into the same loop.
- A single dev box with CarMaker 13.1.2 + VISSIM 2022 installed is sufficient to validate this. No multi-machine setup needed.

yunlishao-vibe:
> ⚠️ **Stacked PR — do not merge before its prerequisites.** This branch was rebased onto the #158 Stage B/B+ work, so the commits below the 35 `#168` commits belong to open PRs **#165 (Stage B)**, **#163 (Stage B+)**, and **#162 (Stage C)**. Merge those into `dev_v0.9.0` first; GitHub then recomputes this diff down to the 35 `#168` commits. The #168-specific change is **55 files, +9888 / −2095**.

> ⚠️ **Before merge — ProprietaryFiles pointer.** The submodule pointer is on a PF *feature* branch (`heads/feature/168_cm_office_xil` @ `2fb6755`), not `PF/main`. Per the ProprietaryFiles workflow in `CLAUDE.md`, the companion PF PR must merge first, then this pointer is fast-forwarded to the new `PF/main` HEAD.

## Summary

Brings **real CarMaker office** into the Stage B DSProxy co-sim loop, replacing the `fake_carmaker.py` stand-in (#168). TrafficLayer (DSProxy mode) drives VISSIM 2022 via `DrivingSimulatorProxy.dll` and exchanges `VehFullData` over TCP with CarMaker's `VirtualEnvironment.lib` (User.c hook): CarMaker pushes ego pose and receives VISSIM traffic state every tick. Ships a click-to-run probe under `tests/Vissim/Probes/TrafficLayer_DSProxy_CMoffice/` — a red Ferrari ego on a smooth junction-authored loop with background traffic — proven end-to-end, GUI and headless.

Key pieces:
- **CarMaker office ↔ FIXS socket bridge** working against the real User.c hook (not the Python equivalent): ego present in VISSIM at the CarMaker-driven position; VISSIM traffic rendered in the CarMaker scene.
- **Smooth drivable loop** — `netconvert` can't emit rounded corners, so the loop `.xodr` is authored junction-based (`gen_junction_loop_xodr.py`); IPGDriver navigates it via a generated Route.
- **Graceful co-sim shutdown** — TrafficLayer detects peer close (recv EOF) on CarMaker stop and ends the run `SIM_END` (clean), not `SIM_ABORT`; no zombie processes.
- **Run length owned by config** — `config.yaml` `SimulationEndTime` drives TrafficLayer's tick budget; the CarMaker maneuver is a 9999 s backstop.
- **Cheap traffic at scale (`UpdRate=200`)** — re-enables the geometry-cost lever retracted in `0c6e8d57`, now that the phase-lock freeze is understood and fixed (Root Cause #2): **12.7× cheaper** CarMaker-core traffic (367 → 29 µs/step), real-time at higher object counts, ego trace identical to the 1000 baseline.
- **Notable bug fixes — two distinct traffic freezes** (see Root Cause sections; FP-gate audit in #169).

## Root Cause #1 — the refresh-gate freeze (`56a87ca7`)

The `.lib` traffic refresh/interpolation gate fired on `fabs(simTime*refreshRate - round) < 1e-5`. `simTime` accumulates FP rounding error that grows with run length; at `refreshRate = 1000` (1 ms refresh) that error exceeded `1e-5` **permanently at ~715 s**, after which the gate stopped firing and **all** traffic-vehicle interpolation into CarMaker froze (the two sims drifted ~190 m apart by 1000 s, neither erroring). Fixed by replacing the fixed-tolerance test with an exact integer-slot edge detector (`llround`), which never drifts and fires once per slot regardless of run length. Verified: a full 1000 s headless run holds a flat offset. **#169** tracks the sibling instances of this construct elsewhere in the codebase.

## Root Cause #2 — the UpdRate phase-lock freeze (`d506e1e3`)

Lowering CarMaker's traffic `UpdRate` below the 1 kHz sim rate (to bank cheaper geometry) froze the teleported FreeMotion traffic — the reason `0c6e8d57` retracted the 200 lever. This is a **separate** mechanism from Root Cause #1. CarMaker reconciles the externally-written `t_0` against its own FreeMotion model at `UpdRate`; when the `.lib` throttled its `t_0` writes to a rate **matched or harmonically related** to `UpdRate` (`TrafficRefreshRate` = 0.5× / 1× / 2× `UpdRate`), the reconciliation repeatedly saw Δ≈0 and discarded the write, resetting each car to CarMaker's own value. Freeze requires BOTH clocks below the sim rate AND harmonically locked — exactly why the 1000/1000 default never exhibited it (both ride the sim ceiling; every step is sampled). Confirmed independently three ways (per-step `t_0` trace, `.lib` re-park readback, write-rate sweep) and shown **not** to be a hook-order artifact. Consistent with CarMaker's documented design: `FreeMotion=1` defaults to `UpdRate=1000` and external-input maneuvers force every-step updates.

**Fix (two parts):** (1) keep `TrafficRefreshRate = 0.001` so the `.lib` writes `t_0` every sim step — CarMaker then always samples a fresh, non-aliased position regardless of `UpdRate`, decoupling the write rate from the geometry-update rate; (2) re-park unmapped RS_C slots (`t_0[2] = -5000`) every refresh so they HOLD buried at low `UpdRate` instead of drifting onto the road and blocking the ego. Net: `UpdRate=200` is real-time-safe — **12.7× cheaper** geometry (367 → 29 µs/step) for ~5 µs/step of `.lib` work; ego trace + peak traffic count identical to 1000. Full evidence trail in `PERF_N_TRAFFIC.md`.

## Related Issues / Tasks

- **Closes #168** (real CarMaker office validation of Stage B).
- **Stacks on #165 / #163 / #162** (#158 Stage B / B+ / C — merge first).
- Filed the audit **#169** (FP-equality timing gates that silently freeze over long runs).
- Relates to **#158** (DSProxy umbrella), **#167** (DSProxy absorption — this probe is a regression scenario to preserve), **#117** (real-time pacing, explicitly deferred), **#101 / #156** (original CM-VISSIM sync intent).

## Type of Change
- [x] Bug fix (FP refresh-gate freeze; UpdRate phase-lock freeze; double-ego; graceful shutdown)
- [x] New feature (CarMaker office co-sim probe + socket bridge; UpdRate=200 cost lever)
- [ ] Maintenance / Refactor
- [ ] Documentation
- [x] Test case / scenario update (verify_demo, headless harness, perf characterization scripts)

- **CommonLib** — `VirEnvHelper.cpp` (FP refresh gate + re-park-unmapped-slots UpdRate-freeze fix + RS_DEBUG instrumentation), `VirEnv_Wrapper.cpp` (peer-close → SIM_END), `SocketHelper.cpp`.
- **TrafficLayer** — `DSProxyMode.cpp` (CarMaker co-sim: ego re-stamp, vehicleClass/heading/type, traffic publish).
- **VirtualEnvironment / TrafficLayer `.vcxproj`** — opt-in `RS_DEBUG` build flag.
- **ProprietaryFiles** — submodule pointer bump (CM13 project: ego `.car` CG fix, driver template, TestRun @ `UpdRate=200`, gated User.c diagnostic movers).
- **tests/Vissim/Probes/TrafficLayer_DSProxy_CMoffice/** — probe (README, `build_testrun.py` [`UpdRate=200` default], `config.yaml` [`TrafficRefreshRate=0.001`], network, runners, `verify_demo.py`) + perf characterization (`PERF_N_TRAFFIC.md`, `diag_*.py`).
- **tests/Vissim/SimpleEcho/** + **tests/Python/SimpleEchoClient/** — junction-authored smooth loop geometry (`.xodr` / `.inpx` / SUMO net).

## Test Cases
- **Headless self-test:** `verify_demo.py` — launches TrafficLayer + VISSIM + headless CarMaker and asserts the round-trip over a long run (offset stays flat across the full 1000 s; was ~190 m by ~715 s before the FP fix).
- **UpdRate=200 perf validation:** `diag_cost.py` (12.7× geometry saving, 367 → 29 µs/step), `diag_sync.py` (~7 cm extra staleness, no sync break vs 1000), `diag_mover_hz.py` / `diag_seq.py` / `diag_libseq.py` (phase-lock alias proven three independent ways); narrative in `PERF_N_TRAFFIC.md`.
- **Click-to-run GUI demo:** `run_cm_office_demo.bat` — TrafficLayer spawns VISSIM, the CarMaker office GUI auto-loads the TestRun + IPGMovie; START begins the lockstep co-sim.
- **Headless demo:** `run_demo_headless.bat`.

## Environment
- Python version: 3.10 (probe scripts)
- VISSIM version: 2022
- IPG CarMaker version: 13.1.2 (office; no Simulink, no CM4SL, no dSPACE)
- C++ compiler/version: Visual Studio 2022
- Windows: 11 24H2

## Checklist
- [x] Code compiles/runs as expected (TrafficLayer + VirtualEnvironment.lib + headless CarMaker)
- [x] Tests pass locally (`verify_demo.py` round-trip; GUI demo confirmed; UpdRate=200 perf + sync validated)
- [x] Documentation is updated (probe `README.md`, `PERF_N_TRAFFIC.md`; `doc/CarMakerDoc.md` cross-ref still pending — see notes)

## Additional Notes
- **Merge order (critical):** #165 → #163 → #162 → companion PF PR → reconcile PF pointer to `PF/main` → this PR. Until the stage PRs land, this PR's diff includes their commits.
- Acceptance item "cross-reference from `doc/CarMakerDoc.md`" is **not yet done** — flagging as a small follow-up.
- `RS_DEBUG` instrumentation (traffic-pose / id-map / refresh dumps) is compiled out by default; opt in with `-p:RS_DEBUG=1`. Dumps are gitignored.
- The `UpdRate=200` default banks the cost win for the shipped demo; revert a single `build_testrun` knob (`RS_UPD_RATE=1000`) to fall back to the conservative rate.


cjffly:
When operate in VISSIM, like zoom in/out, the simulation will pause for both side very shortly, Maybe a VISSIM problem?

yunlishao (approved):

yunlishao-vibe:
@cjffly Good catch — it's a VISSIM-side effect, and by design rather than a bug in our pipeline. We instrumented the DSProxy loop per-tick: zooming/panning spikes the time TrafficLayer spends blocked in `VISSIM_GetTrafficVehicles` from ~5–10 ms to 100–450 ms, while the rest of the tick (ego push, signal pull) stays flat — the lockstep waits on VISSIM, so both sides pause briefly.

This is documented in PTV's manual **"PTV Vissim API – Driving Simulator Interface"**, which ships with VISSIM at `…\PTV Vissim 2022\API\DrivingSimulator_DLL\doc\PTV Vissim API - Driving Simulator Interface.pdf`. Two passages:

- §3 *DLL Interface Details* (p.11), `VISSIM_GetTrafficVehicles` — *"Blocks while the calculation of the time step in Vissim hasn't finished yet."* (The same comment is in the API header `…\DrivingSimulator_DLL\include\DrivingSimulatorProxy.h`.)
- §2 *Architecture* (p.6) — *"The simulation speed in Vissim needs to be set to 'maximum', and the visualization in Vissim should be switched off in order to achieve the highest possible simulation speed. The DS can slow down the co-simulation by waiting between DLL calls as much as desired … without hurting the Vissim simulation at all."*

VISSIM runs the sim step and the network window on one thread, so an interactive zoom (input + viewport recompute + redraw) delays step completion and `GetTrafficVehicles` blocks longer. We confirmed it persists with Quick Mode on **and on VISSIM 2026**, with no related fix in the 2022/2026 release notes — so it's architectural, not version-specific. (Quick Mode does cut the *idle* per-step cost ~6 ms → ~2 ms; it just can't offload an interactive zoom, since there's no second thread.)

On the fix direction: rather than asking people not to touch the VISSIM view (we *do* want to watch it) — and pausing CarMaker isn't an option once it runs in XIL/HIL real time — the better approach is to **automate the VISSIM launch so it opens already framed on the ego vehicle and follows it**, so there's no need to manually zoom/pan, which is what triggers the stall. I'm checking whether VISSIM's COM API can set that follow-camera/viewpoint at startup (or whether we bake it into the saved layout). The underlying single-thread blocking stays, but it isn't hit if the view doesn't have to be driven by hand.

Issue #176, opened 2026-06-19
[BUG] YAML 'all' vehicle subscription is dead code (subscribeAllVehicle parsed but never consumed)


yunlishao:
**Describe the bug**
The YAML "subscribe to all vehicles" option is **dead code**. The vehicle-subscription attribute `all` (e.g. `attribute: {all: ['true']}`) is parsed into `ConfigHelper`'s `subscribeAllVehicle` (`std::pair<bool,double>`) at `CommonLib/ConfigHelper.cpp:945` and `:955` and stored on the subscription struct (`ConfigHelper.h:208`, `TrafficHelper.h:93`), **but it is never read anywhere**. A full-repo grep shows `subscribeAllVehicle` is only ever written/passed (`ConfigHelper.cpp:299,305,868,945,955`; `TrafficHelper.cpp:411,415`) — there is no `if (subscribeAllVehicle.first)` branch in the codebase. So `all: ['true']` is a silent no-op.

As a result, on the SUMO path the only working visibility lever is the per-ego **`radius`** context subscription (`CommonLib/TrafficHelper.cpp:1049`, `Vehicle::subscribeContext(id, CMD_GET_VEHICLE_VARIABLE, radius, ...)`). To "send all vehicles" you must set a large `radius`; the documented `all` identifier (README / CLAUDE.md describe an "All" subscription) does nothing.

**Related divergence (context, surfaced under #174):** the DSProxy/VISSIM path (`TrafficLayer/TrafficLayer/DSProxyMode.cpp` ~L411–435) hardcodes publish-all and ignores the YAML subscription entirely (`VisibilityRadius: -1`). So the *same scenario shows a different vehicle set per backend* — VISSIM-CM ≈ all (~40 on SimpleLoop), SUMO-CM ≈ only vehicles within the 100 m radius. The fix should make the displayed set config-driven and consistent across backends.

**To Reproduce**
Using `tests/Python/SimpleEchoClient` (SUMO + TrafficLayer + a pure-Python subscriber; no CarMaker/VISSIM needed):
1. Run the harness with `ApplicationSetup.VehicleSubscription` = `attribute: {id: ['ego'], radius: [0]}` and note the received vehicle count (echo client prints `Vehicles: N`).
2. Change the subscription to `attribute: {all: ['true']}` and re-run.
3. **Expected:** all vehicles in the network are received. **Actual:** the count is unchanged by the `all` flag (the flag is never consumed); only changing `radius` changes the count.

[...]

Issue #130, opened 2026-03-13
[MAINTENANCE] Python CommonLib: add package structure, fix imports, and expand Python test suite


yunlishao-vibe:
## Task Summary

The Python CommonLib (`ConfigHelper.py`, `MsgHelper.py`, `SocketHelper.py`, `VehDataMsgDefs.py`) lacks a proper package structure, has inconsistent import patterns, and the Python test suite under `tests/Python/` needs expansion to cover core message and socket functionality.

## Why is this needed?

Currently:
- No `__init__.py` in `CommonLib/` — it's not a proper Python package, imports rely on `sys.path.insert()` hacks
- Both `tests/Python/SimpleEchoClient/simple_echo_client.py` and `tests/Python/SimpleTrafficLight/echo_client.py` duplicate the `sys.path.insert` pattern manually
- `SocketHelper.py` has a `# TODO: add traffic light and detector data` comment
- Python tests only validate socket echo behavior — no unit tests for message packing/unpacking
- `MsgHelper.py` at 28K lines is likely the largest file in the Python codebase and has never been unit tested
- `requirements.txt` references outdated SUMO version numbers in comments and is inconsistent with `environment.yml`
- Two separate conda environment definitions (`environment.yml` and `requirements.txt`) with different package lists

### 1. Add Python package structure
- Add `CommonLib/__init__.py` exposing key classes
- Update all test files to use clean `from CommonLib.X import Y` imports
- Remove sys.path hacks from test scripts

### 2. Fix requirements files
- Reconcile `requirements.txt` and `environment.yml` — one should be authoritative
- Remove outdated version comments in `requirements.txt` (e.g., `vissim==11.0`, `carmaker==10.0.1`)
- Ensure `environment.yml` includes all packages needed to run Python tests without SUMO/VISSIM/CarMaker installed

### 3. Implement SocketHelper.py TODO
- The `# TODO: add traffic light and detector data` in `SocketHelper.py` — implement or convert to proper issue

### 4. Expand Python unit tests
Add `tests/Python/unit/` directory with:
- `test_msg_helper.py` — test message packing/unpacking with known byte sequences
- `test_config_helper.py` — test YAML config parsing with sample config files
- `test_socket_helper.py` — test message header parsing (no live socket needed, use mock data)

### 5. Add CI-runnable test runner
- Add `tests/Python/run_unit_tests.py` (or pytest config) that runs all unit tests without requiring any simulator
[...]

PR #220, opened 2026-07-29
#87/reliable TCP framing for large messages across cpp, python and simulink

### LINKED ISSUE #87: [FEATURE] Support large message sizes with end-of-FIXS message marker in TCP/IP
Currently, the TCP/IP communication in RS_FIXS may have limitations handling large messages. Without a proper end-of-message delimiter or marker, it's difficult to reliably transmit and receive variable-length messages, especially when dealing with large vehicle data sets or complex traffic scenarios that exceed typical buffer sizes.

Implement support for large message sizes in the TCP/IP communication layer with a well-defined end-of-FIXS message marker/delimiter. This would include:

1. Define a unique end-of-message marker (e.g., a specific byte sequence like `\r\n\r\n`, `\xFF\xFF\xFF\xFF`, or a custom delimiter)
2. Update the `SocketHelper` class to handle message framing with this delimiter
3. Implement proper buffering to accumulate partial messages until the delimiter is received
4. Handle edge cases where the delimiter might be split across multiple TCP packets
5. Add support for messages larger than the current buffer size by dynamically growing buffers or using chunked reads
6. Ensure backward compatibility with existing message formats

- **Length-prefix approach**: Prepend each message with a fixed-size header containing the message length. This requires knowing the full message size upfront.
- **Fixed-size messages**: Pad all messages to a maximum size, which is wasteful for smaller messages.
- **Custom binary protocol**: Implement a more complex protocol with message types, lengths, and checksums, though this adds significant complexity.
- **Using existing serialization libraries**: Adopt Protocol Buffers, MessagePack, or similar, though this would require larger refactoring.

- Python version: N/A
- C++ compiler/version: (relevant for SocketHelper implementation)
- MATLAB/Simulink/dSPACE version: N/A
- CARLA version: N/A
- SUMO version: (impacts message sizes)
- VISSIM version: (impacts message sizes)
- IPG Carmaker version: N/A

**Additional context**  
Looking at TrafficHelper.cpp, the system handles vehicle data through SocketHelper when using VISSIM mode. Large scenarios with many vehicles or extensive subscription data could exceed current buffer limits.

Key files to modify:
- CommonLib/SocketHelper.cpp / CommonLib/SocketHelper.h - Core TCP/IP communication
- CommonLib/MsgHelper.cpp / CommonLib/MsgHelper.h - Message packing/unpacking
- CommonLib/TrafficHelper.cpp - Integration with traffic simulators

Example pseudocode for the delimiter approach:
\`\`\`cpp
const std::string END_OF_MESSAGE = "\r\n\r\n";

// Sending side
void sendMessage(int socket, const std::string& message) {
    std::string messageWithDelimiter = message + END_OF_MESSAGE;
    send(socket, messageWithDelimiter.c_str(), messageWithDelimiter.size(), 0);
}

// Receiving side
std::string receiveMessage(int socket) {
    std::string buffer;
    char chunk[1024];
    while (true) {
        int bytesRead = recv(socket, chunk, sizeof(chunk), 0);
        buffer.append(chunk, bytesRead);
        
        size_t delimiterPos = buffer.find(END_OF_MESSAGE);
        if (delimiterPos != std::string::npos) {
            std::string message = buffer.substr(0, delimiterPos);
            // Keep remaining data for next message
            buffer = buffer.substr(delimiterPos + END_OF_MESSAGE.length());
            return message;
        }
    }
}
\`\`\`

This enhancement would improve robustness and scalability of the RS_FIXS communication layer.

yunlishao-vibe:
## Summary

Makes the FIXS TCP layer safe for large messages, consistently across **all four** implementations of the wire protocol (C++, Python, Simulink/dSPACE MEX, and the test harness). The message format itself is **unchanged**, so a fixed peer interoperates with an unfixed one and the fixes can land in any order.

Scoped to the transport layer. It does **not** remove `N_MAX_VEH` / the `min(…, 200)` edge cap — see *Deliberately out of scope*.

### One correction to the issue's premise

The issue proposes an end-of-message delimiter. The protocol is already **length-prefixed** — a 9-byte header (`uint8 simState`, `float simTime`, `uint32 total_msg_size`) then per record `uint16 size` + `uint8 type` — so a length prefix plus an end marker would be redundant, and changing the format would force a flag day across every client. `uint32` allows 4 GB, which is not the constraint. Rescoped to *reliable framing on the existing length prefix*.

An end marker would only be needed if the sender could not know the total up front (it can: the send lists are fully populated), if one timestep were fragmented across several headered messages, or as a resync point after a desync — and that last one papers over the bug instead of fixing it.

### What was wrong

**Short reads desync the stream permanently (C++ + Python).** `recv()` returns only what has arrived. Every read site assumed it returned the full request. Because the parse loop advances by the *declared* record size, the unread tail of one record becomes the next record's header, and the connection never recovers. This is invisible with a handful of vehicles (one segment) and routine under `all` subscription.

**Send-side heap overflow past ~500 vehicles (C++).** `packVehData` takes no capacity and never bounds-checked; every record was packed into a `new char[65536]`. Reachable today: `N_MAX_VEH` only caps the `ENABLE_VEH_SIMULATOR` (CarMaker) path, so a viz-only `all` run has nothing bounding it. Silent heap corruption, not an error.

**~160 KB leaked per `sendData` call.** Four `new char[]` and not one `delete` anywhere in `SocketHelper.cpp` — ~1.6 MB/s at a 10 Hz step.

**`sendBuffer` was 9 bytes short.** `new char[81728]`, but max content is `9 + 65536 + 8096 + 8096 = 81737`.

**Record body could overflow a 1024-byte stack buffer.** Worst-case vehicle record is **1888 B** (`3 + 7 strings × (1+255) + 93` numeric); the size went straight to `recv()` unchecked.

**The Simulink MEX never parsed the framing at all.** One `recv()` of ≤1024 B handed raw bytes to the model: ~7 vehicles max, and the remainder of any longer message stayed in the socket, so every later step parsed mid-record.

**The regression test could not fail.** `_ReplaySock` always satisfied the full request, so the receive loop was never exercised — which is why this survived the #176 fix that introduced the file.

### How it's fixed

- `recvExact` (C++) / `_recv_exact` (Python) / `recvExactMex` (MEX) — loop until N bytes arrive. The C++ version returns `0` only for a clean close *before any byte*, preserving the "peer closed gracefully" branch `DSProxyMode` uses to tear VISSIM down when CarMaker stops; a mid-read close is truncation and reports `SOCKET_ERROR`.
- **Receive needs no large buffer.** `total_msg_size` is a *loop bound*, not an allocation size — records are parsed one at a time into `VehDataRecv_um`, so a 10 MB message is consumed with a 2 KB buffer. `MAX_RECORD_SIZE = 8192` is the single record ceiling, enforced on **both** emit and accept so the two ends agree; an out-of-range size fails loudly. The 1888 B bound is provable only for *vehicle* records — a detector record carries a `vector<DetectorData_t>` and has no ceiling below the `uint16` 65535.
- **Send streams in chunks**: a size pass (so the header can go out first), then pack-and-flush through one reused `TX_CHUNK_SIZE` member buffer, with `sendExact` for partial writes. Message size is unbounded and the buffers are fixed — no per-call `new`, which is what the dSPACE/CarMaker real-time targets need. (Not literally allocation-free end to end: `sendData` still takes `MsgHelper` by value, pre-existing and noted below.) This removes the overflow, the leak, and the off-by-9 in one change.
- **No drift between sizing and packing**: a `NULL` buffer means *measure, don't write* — `vehRecordSize`/`tlsRecordSize`/`detRecordSize` run the same `pack*()` traversal, so adding a message field keeps them in sync automatically. (This replaces the parallel size function I originally planned.)
- **MEX: drain ≠ deliver.** It now always consumes exactly `total_msg_size`, delivers what the fixed budget holds, and **discards the rest** so the stream stays aligned. Dropping records costs no memory, which is what makes an unbounded message safe on a target with static storage — the cap bounds what is *delivered*, never what can be *received*. Overflow is logged with byte counts.
- **The budget stays in bytes** — and stays at **1024**, see below — not a vehicle count: record size isn't knowable at compile time (it depends on the configured `VehicleMessageField` set and on runtime id/link/lane string lengths) while the output port and DWork widths are literally byte counts. Same unit as the constraint means no conversion constant that goes stale when the YAML changes.
- **The MEX delivery budget is deliberately NOT raised.** `RECVBUFFERSIZE`/`SENDBUFFERSIZE` set the S-function port widths, so changing them breaks every existing Simulink model on load. The #87 defect is the desync, which drain-and-discard fixes without touching the interface — so this PR is **non-breaking**. Models keep loading and get a *correct* ~7 vehicles instead of a corrupted stream. Raising the budget belongs in its own change with model migration.

### Incidental fixes in touched code

- `TCP_NODELAY` was commented out in the MEX path while TrafficLayer sets it. This matters more now the peer streams in chunks: Nagle holds a sub-MSS write until the previous segment is ACKed, and against delayed ACK that's a 40–200 ms stall — fatal to a 10 Hz lockstep.
- `memset(recvDataHold, 0, sizeof(recvDataHold))` used the **pointer** size (8 B), not the DWork size, so the buffer was never really cleared. Fixed at all three sites, using `RECVBUFFERSIZE` — DWork 0 is `SS_INT8`, so its width *is* its byte count. (An earlier revision of this branch used `RECVBUFFERSIZE * sizeof(real_T)`, which overran the DWork by 8x; caught in review, see `89788116`.)
- The MEX first-step handshake read also grabbed one bufferful and dropped it, desyncing from step one; it now consumes exactly one message.

## Related Issues / Tasks

Closes #87. Builds on the framing work in #176 (`f4c8a9d3` fixed `total_msg_size` on the **Python** side only; the C++ side and short-read robustness were untouched).

Supersedes `feature/#87`, which is 267 commits behind and whose diff is ~80% a duplicate subscribe-all implementation that landed independently as #176. **That branch can be deleted.**

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Maintenance / Refactor
- [ ] Documentation

- `CommonLib/SocketHelper.cpp` / `.h` — `recvExact`/`sendExact`, chunked `sendData`, `MAX_RECORD_SIZE`, `TX_CHUNK_SIZE`
- `CommonLib/MsgHelper.cpp` / `.h` — measure-only pack pass, `vehRecordSize`/`tlsRecordSize`/`detRecordSize`
- `CommonLib/SocketHelper.py` — `_recv_exact`
- `CommonLib/RealSimSocket.cpp` — framing loop, byte budget, `TCP_NODELAY`, memset fix
- `tests/Python/SimpleEchoClient/test_msg_framing.py` — `_ChunkedReplaySock`, `test_framing()` pytest entry point
- `TrafficLayer/x64/Release/TrafficLayer.exe` — rebuilt (see checklist note)

## Test Cases

**Red → green, verified in that order.** The test harness commit landed first specifically so the failure was observed before any production change:

```
# before the fixes
FAIL  (200,200)   recv<=1B: recv_data raised error: unpack requires a buffer of 4 bytes (stream desync)
FAIL  (200,200)   recv<=7B: ...
FAIL  (1000,1000) recv<=1B: ...
FAIL  (1000,1000) recv<=7B: ...
4 case(s) FAILED
```

`k=64`/`k=1500` pass even unfixed — records in that field set are ~40–50 B, so a short read only occurs once the chunk is smaller than a single request. Same reason the bug is invisible with few vehicles and routine under `all`.

```
<realsim_python> tests/Python/SimpleEchoClient/test_msg_framing.py
  -> All framing cases passed (#176 regression + #87 short-read/large-message)
     11 cases: N up to 1000 x recv chunk in {1, 7, 64, 1500}

<realsim_python> -m pytest tests/Python/unit/ -q
  -> 34 passed

MSBuild TrafficLayer.vcxproj -p:Configuration=Release -p:Platform=x64
  -> TrafficLayer.exe built; no new warnings from SocketHelper.cpp / MsgHelper.cpp

mex -c RealSimSocket.cpp   (R2024a)
  -> MEX completed successfully
```

**End-to-end C++ → Python** (the cross-language check that matters — a wrong `total_msg_size` from the new size pass would desync the Python client immediately):

```
<realsim_python> tests/Python/SimpleEchoClient/run_subscription_compare.py

variant       max_vehicles  distinct_total
ego_only                 1               1
radius                  11              11
all                     29              29
-> all > radius > ego_only, stream consumed exactly, no desync
```

## Environment

- Python version: 3.10 (`realsim` conda env)
- MATLAB/Simulink/dSPACE version: R2024a (mex compile only)
- SUMO version: 1.22.0 (libsumo)
- VISSIM version: not exercised
- IPG CarMaker version: not exercised

## Checklist

- [x] Tests pass locally

### Reviewer notes

0. **Review pass applied** (`89788116`): a DWork memset overrun this branch introduced, a send/receive record-cap disagreement, an inconsistent close check on the body read, and by-value copies in the new size functions. See the PR comments for the full accounting.
1. **Not validated on real hardware.** The MEX compiles and the framing logic is exercised by the Python tests, but no MATLAB/Simulink model run and no dSPACE target build were performed. The delivery-budget and discard paths should get an in-model check before release. This is the main residual risk.
2. **VISSIM and CarMaker paths not exercised.** `recvData` is shared with `DSProxyMode`; the graceful-close semantics are preserved by construction (documented in `recvExact`) but not run.
3. **`TrafficLayer.exe` is committed.** Tracked, and recent feature commits refresh it (`197516af`, `082566f5`), so I kept it consistent rather than shipping an exe predating the fix — the SimpleEchoClient harness and the fetched release bundle both run it. Force-added past a directory-level gitignore. **Drop `9217cca6` if you'd rather the v0.9.0 CI produce it.**
4. **`sendData` still takes `MsgHelper` by value**, copying every send map on each call. Pre-existing and unrelated; left alone to keep the diff reviewable, but worth its own issue.

### Deliberately out of scope

- **Cap removal** (`N_MAX_VEH = 100`, `min(…, 200)`). Neither is reachable from a viz-only `all` run, and a cap at the CarMaker boundary is arguably correct given finite traffic objects — the defect is that it's hardcoded and silent, not that it exists. Better as a config key (`MaxVehiclesSent`, `-1` = unlimited) plus a log line on truncation.
- **Nearest-first send ordering.** Would make byte-budget truncation semantically meaningful anywhere downstream (currently "whatever fits" is `unordered_map` order — arbitrary and unstable between steps). Real improvement, but a server-side behavior change rather than a framing fix.


yunlishao-vibe:
Pushed `d9919b1b` — a real overflow I introduced in `aa156bd1`, found while explaining `tlsRecordSize`.

The send-side flush guard reserved `MAX_RECORD_SIZE` per record. That constant is the **vehicle** worst case (1888 B), but a detector record isn't bounded by it: `TlsDetector_t` is `tuple<int, string, vector<DetectorData_t>>`, so its size grows with the detector count at an intersection and is capped only by the `uint16` size field (65535).

```
record = 3 + (1 + name) + 2 + nDet x (1 + detName + 1 + 1)
```

Exceeded at ~7 max-length detector names, or ~77 realistically-named ones. The guard allowed `txLen` up to `TX_CHUNK_SIZE - 2048`, so packing a 3000-byte detector record wrote past `txBuf` — the same class of overflow this branch exists to remove, just relocated.

Fix: reserve the record's **actual** size (the `*RecordSize` functions already exist for the header size pass) and refuse a single record larger than `TX_CHUNK_SIZE` loudly instead of overrunning. Also `MAX_RECORD_SIZE` 2048→8192 with an honest comment (the provable ceiling only holds for vehicles), `TX_CHUNK_SIZE` 16384→32768, and removed a duplicate define.

Re-verified after the change: framing suite green, 34 unit tests, TrafficLayer rebuilt with no new warnings, and end-to-end C++→Python still `all=29 > radius=11 > ego_only=1`.

Worth noting for review: **the detector path is the least-tested part of this PR.** The Python receiver still no-ops on detector records (`# see #130`), so nothing here exercises a large detector record end-to-end. The bound is now enforced rather than assumed, but a detector-heavy scenario would be the thing to try if one exists.

yunlishao-vibe:
Pushed `89788116`. Review findings verified individually rather than accepted wholesale — most were right, one was not, and on one I went further than suggested.

**#1 memset — confirmed, my bug, and the worst thing in the branch.** `ssSetDWorkDataType(S, 0, SS_INT8)` with width `RECVBUFFERSIZE` means the DWork is 16384 **bytes**; `RECVBUFFERSIZE * sizeof(real_T)` wrote 131072 — a 114 KB overrun per timestep on the real-time target. The bug I replaced under-cleared (harmless); mine corrupted. Fixed at all **three** sites — `mdlStart:389` still had the original, so two of three had been touched and both touched wrongly. Pointer retyped `int8_T*` to match the declared type.

**#2 cap disagreement — confirmed.** `MAX_RECORD_SIZE` is now the single wire contract, refused on emit *and* on accept; `TX_CHUNK_SIZE` is purely a syscall-batching knob. Python enforces the same ceiling — not for symmetry (no sender emits one now) but as desync detection: on a bogus size Python would block forever where C++ fails loudly.

**#3 close handling — confirmed.** Body read now `<= 0`, guarded so a legitimate zero-length body doesn't trip it.

**#4 by-value — confirmed**, and it directly contradicted the "allocation-free" claim. All three are `const&`. `packVehData` itself is still by-value: pre-existing, and widening the diff into a shared signature isn't worth it here.

**#5 txBuf — confirmed.** Comment corrected; it claimed a stack benefit it doesn't deliver, since `SocketHelper` is a stack object at three call sites.

**#6 — I went further than the suggestion.** Rather than *document* the port-width break, I reverted `RECVBUFFERSIZE`/`SENDBUFFERSIZE` to 1024. The #87 defect is the desync, not the buffer size, and drain-and-discard fixes it without touching the interface. This PR is now **non-breaking**: existing models load unchanged and get a correct ~7 vehicles instead of a corrupted stream. Raising the budget deserves its own change with model migration.

**#7 — confirmed and fixed.** `test_msg_framing.py` had no `test_*` function, so pytest collected nothing. Added `test_framing()`; script mode kept.

**Where I disagree: the `#include <CarMaker.h>` item is not in this diff.** That NOTE and the missing include are already in `origin/dev_v0.9.0`:
```
git diff origin/dev_v0.9.0..HEAD -- CommonLib/SocketHelper.h | grep -E '^[-+].*include'   # no output
```
Pre-existing state, misattributed.

**Not fixed, deliberately:** the 4-byte handshake `recv`/`send` at SocketHelper.cpp:749/:800. Real inconsistency, but it's a control transfer on a fresh socket outside the framed-data path, and using the helpers there means relocating them for negligible gain. Noted as follow-up. `VirEnvCore.cpp:268` likewise — pre-existing, <2 KB payloads, and it's a fifth component rather than one of the four message implementations.

PR body updated for the constant drift (2048 → 8192) and the reverted budget.

Re-verified after all of it: pytest collects the framing test (1 passed), 34 unit tests, clean TrafficLayer rebuild, R2024a mex compile, end-to-end C++→Python unchanged at all=29 > radius=11 > ego_only=1.

**On the binary (`9217cca6`):** fair point that it's unreviewable, and it has now been rebuilt twice mid-review. Happy to drop it and let CI own it — say the word.

yunlishao-vibe:
All four residual items fixed in `f0a9cddb`, plus a merge that turned out to be necessary — `d2362a92`.

1. **Comment drift** above `FIXS_TX_FLUSH_IF_NEEDED` — rewritten to describe the contract actually enforced. Same drift class as the 2048/8192 one; fair catch twice.
2. **#4 was half-landed — correct.** `const&` removed only the outer copy; the `pack*` functions were still by value, so the inner copy fired on every measure call. All three `pack*` signatures now take `const&` (every call site passes an lvalue, so source-compatible). That's what actually closes it.
3. **The per-step printf — best catch of the round**, and a direct consequence of my revert. At a 1024 B budget an `all` run overflows every step, so an unconditional `printf` in `mdlUpdate` is a per-step console write on the RT target — the same class of stall the `TCP_NODELAY` fix in that very file removes. Latched to once per run via `gOverflowWarned`, reset in `mdlStart`; `enabledebug` still reports every step.
4. **Stale `recvSize`** — `recvSize = 0` assigned unconditionally before the short-circuit.

**Binary dropped**, as agreed. Three rebuilds mid-review is the argument, made empirically.

---

**The merge was not optional.** `dev_v0.9.0` had advanced 96679e64 → db0a81ee and the PR was `CONFLICTING`. #218 added `steerAngleDesired`/`acceleratorPedalDesired`/`brakePedalDesired` at the *"add new vehicle message field here"* marker in `packVehData` — precisely where this branch changed the surrounding lines.

That collision happens to test the branch's central design decision, so it's worth recording: **those three fields needed no change to the sizing code.** They flow through `numericVehDataToBuffer`, which carries the `if (buffer)` guard, so `vehRecordSize` picks them up from the same traversal `packVehData` uses. A parallel size function — the obvious implementation, and the one originally planned here — would have silently under-counted `total_msg_size` by 12 bytes per vehicle on this merge. That is exactly the #176 bug returning, one merge later.

I added the three fields to the framing test's `FIELDS` so the property is guarded rather than argued: if sizing and packing ever diverge, the test desyncs.

**Status:** `MERGEABLE`, 7 files, no binary. 35 pytest tests, clean TrafficLayer rebuild, R2024a mex compile, end-to-end C++→Python at all=29 > radius=11 > ego_only=1 — all re-run after the merge.

**Unchanged recommendation: hold for an in-model MEX run.** Three of the five defects found across both review rounds were in `RealSimSocket.cpp`, the one file with no test coverage. That ratio is the argument.

yunlishao (approved):

Issue #191, opened 2026-07-23
[MAINTENANCE] Automated single-canonical release CI (source-visibility split) + FIXS_VERSION/compat fixes


yunlishao:
**Task Summary**

Stand up an automated GitHub Actions release pipeline that rebuilds and republishes the rolling `latest.zip` on every push to `main`, producing **one canonical release zip for all consumers** (no tiered/diverged releases). Fold in the long-broken version-stamping fixes.

This is the concrete design for #57 (Auto CI process for build). Related: #131 (version tracking), #128 (RealSim→FIXS branding), #190 (manual repack this eliminates).

**Why is this needed?**

- Today the release is fully manual on a developer box: `dispatch.bat` → `build/` → `scripts/dispatch/8_create_zip.ps1` → `scripts/dispatch/publish_release.ps1` (`gh release create latest`). Every downstream-facing change — even a one-line Carla/Python edit — requires a developer to hand-run the full build and republish (see #190).
- Downstream consumers (e.g. FIXS_Applications' `initialize.sh`) pull the `latest` release. Automating it means Python/Carla and core changes land for consumers with no manual repackaging.
- Keeping **one** canonical zip (rather than a lightweight "apps" zip split from a "full" zip) avoids version divergence and the "which zip do I need" problem — every consumer always gets the complete, consistent artifact.
- FIXS is a public repo, so GitHub-hosted Actions runners are free with unlimited minutes (Windows included — the private-repo 2× Windows multiplier and 2,000-min cap do not apply). The automated build has no minute-budget concern, so "always rebuild, even for a trivial change" is effectively free.

**Scope of Work**

### 1. Build split by source visibility

Draw the automation boundary at **"is the source public?"** so the hosted CI never needs the private submodule or any cross-repo token:

- **Public-source → built automatically by CI** (`windows-latest`, every push to `main`): yaml-cpp (dispatch step 1), `TrafficLayer.exe` / `CoordMerge` (step 2), `VirtualEnvironment` (step 4), and all `Carla/` + `CommonLib` Python. No submodule checkout, no token.
- **Private-source → built manually on a licensed workstation** via the existing dispatch flow (step 3 VISSIM DriverModel DLLs, 5a CarMaker, 5b dSPACE, 6 MATLAB MEX). These require installed proprietary toolchains, so they cannot run on a hosted runner. Packaged as a separate `fixs-proprietary-<sha>.zip`. Rare — only when proprietary code changes.

[...]

Issue #174, opened 2026-06-19
[MAINTENANCE] Consolidate VirtualEnvironment into a backend-agnostic core (CarMaker + Carla) with a cross-simulator test suite


yunlishao:
**Task Summary**

Umbrella refactor (staged across multiple PRs). Today the virtual-environment bridges are **two parallel, incompatible implementations** of the same logic. Consolidate them into a single **backend-agnostic orchestration core** reused by both CarMaker and Carla, with backend-specific behavior supplied through a verb interface — and stand up a **cross-simulator test suite** (seeded from #168) that drives every backend through the same pipeline.

- **CarMaker:** `VirtualEnvironment.lib` (`CommonLib/VirEnvHelper.{h,cpp}` + `VirEnv_Wrapper`) — a static lib compiled against `CarMaker.h`, linked into `CarMakerWin64.exe`, driven per-tick by the CM project's `User.c` via `VirEnv_runStep(SimCore.Time)`. CarMaker owns the loop; ego dynamics live on the CarMaker side; ego flows **out** to FIXS Core.
- **Carla:** `VirCarlaEnv/mainVirCarla.cpp` — a standalone `.exe` that re-implements the same orchestration inline in `main()`, owning its own loop and ticking the Carla world. It shares only CommonLib (`SocketHelper`/`MsgHelper`/`ConfigHelper`/`BridgeHelper`); it does **not** use `VirtualEnvironment.lib` and cannot link it as-is (the lib drags in `CarMaker.h`).

Both perform the same seven-step per-tick skeleton: recv traffic+TLS from FIXS Core → map SUMO id → backend handle → spawn/despawn → set vehicle poses + lights → sync traffic lights → ego readback/teleport → send to FIXS Core. The only real differences are (a) who owns the loop and (b) the backend "verbs" (CarMaker slot-pool + `TrfObj->t_0[]` vs Carla spawn/`SetTransform`).

**Why is this needed?**

- **Eliminate duplication / divergence risk:** two copies of the same bridge logic drift apart (e.g. the Carla path lacks the temporal interpolation the CarMaker path has).
- **Make a Carla/CarMaker-shared core actually possible:** the current lib is CarMaker-bound, so Carla can't reuse it. Lifting `CarMaker.h` out behind an interface is what unblocks a single source of truth.
- **Enable seamless new combinations:** once the core is backend-agnostic, **VISSIM-Carla** becomes essentially free (today only SUMO-Carla exists).
[...]

Issue #137, opened 2026-03-13
[DOCS] ReadTheDocs: fix TOC/autodoc/stale refs + auto-build on main


yunlishao-vibe:
## Task Summary

Wire up Python autodoc, fix broken/orphaned TOC entries, add missing BUILD.md to navigation, update stale file references in DEVELOPER_GUIDE, stub out thin placeholder pages with real content, **and set up automatic RTD builds when `main` is updated.**

## Branch and PR Target
- All PRs for this issue should target the `dev` branch

### 1. Fix index.rst TOC
- Add `BUILD` to the Development section (currently not linked despite being the most useful developer reference)
- **Fix broken `mdinclude`**: `index.rst` starts with `.. mdinclude:: README.md`, but there is **no `doc/README.md`** — the include target is missing, so the landing page intro either silently drops or errors depending on the resolver. Point it at the repo-root README (`../README.md`) or add a `doc/README.md`.
- ~~Orphaned pages~~ — stale: every page under `doc/` is now in a `toctree` (the Carla build guides sit under *Building from Source*); Sphinx reports no orphans.
- Verify all files listed in `index.rst` actually exist
- Note: `CARLAdoc.md` **is** correctly linked (as `CARLAdoc`); an earlier internal note calling it orphaned was stale.

### 2. Fix stale references in DEVELOPER_GUIDE.md
- References `RS_CM_BuildConfig_r2021b.py`, `RS_CM_BuildConfig_r2019b.py` — these no longer exist (the BuildConfig files are now generated by `5b_carmaker_dspace.ps1`). `RealSimCarMakerSetup.py` does still exist.
- Update to reflect current CarMaker build scripts (`RS_CM11_1_2_BuildConfig_2024a.py`, `RS_CM13_1_3_BuildConfig_2024a.py`, `patch_user_c.ps1`)
- ~~CommonLib removals~~ — stale: `MathLibrary.h` is still in the repo (listed as legacy), and the guide never mentioned `RealSimCmHelper` / `ryml_all.hpp`.

### 3. Wire Python autodoc
Now that `CommonLib/__init__.py` exists (PR #133), autodoc can generate real API pages:
- Add docstrings to `CommonLib/ConfigHelper.py`, `MsgHelper.py`, `SocketHelper.py`, `VehDataMsgDefs.py`
- Replace `doc/api/index.md` stub with proper `automodule` directives
[...]

Issue #65, opened 2025-10-16
[FEATURE] Enable both Windows and Linux compile — feasibility audit + open design questions


yunlishao:
> **Status: scoping / feasibility report. No design is locked in.**
> Audited baseline: `origin/dev_v0.9.0` @ `db0a81ee` (*not* `main` — `main` is 138 commits behind and gives a materially different answer, see [Appendix A](#appendix-a--why-the-baseline-matters)).
> Effort figures below are estimates with their drivers stated; the dependency questions in [Open design questions](#open-design-questions) are unresolved and are the real cost centre.

## Task summary

Make **`TrafficLayer`** and **`VirCarlaEnv`** compile (and link) on Linux, and decide how that is built, versioned across distros, and gated in CI.

This issue now carries the feasibility audit for those **two targets only**. The umbrella scope of "Windows *and* Linux compile" is broader — see [Not audited](#not-audited-still-in-this-issues-umbrella).

## Why this is needed

- Removes the hard MSBuild/VS2022 coupling for the two components that have no intrinsic Windows dependency (SUMO co-sim + Carla client are both cross-platform).
- Unblocks Linux-native co-sim for collaborators without a Windows/VISSIM licence.
- Prerequisite for a compile gate in CI on anything other than `windows-2022` (#57, #40).

## Bottom line

| Target | Source-edit effort | Dominated by |
|---|---|---|
| `TrafficLayer` (SUMO-only) | **~1–2 days** | mechanical; mostly filling in `#else` branches that already exist |
| `VirCarlaEnv` | **~3–5 days** | **not** source edits (~1 h) — entirely `libcarla` client toolchain on Linux |

Confidence on `TrafficLayer` is reasonable (the Win32 surface is enumerated below). Confidence on `VirCarlaEnv` is **low until two unknowns are settled empirically** (Q1, Q2) — both are answerable by a build spike, not by more reading.

---

## Evidence — Win32 surface per translation unit

Counted on `origin/dev_v0.9.0` over the exact TU sets in `TrafficLayer.vcxproj` and `VirCarlaEnv.vcxproj`. `guards` = number of existing `#ifdef WIN32` / `#ifndef WIN32` sites, i.e. work already done.

### `TrafficLayer` — 8 TUs (was 5 on `main`)

| TU | Win32 hits | guards | Note |
|---|---:|---:|---|
| `CommonLib/SocketHelper.cpp` | 60 | 23 | already fully dual-pathed |
[...]

PR #229, opened 2026-07-30
#65/TrafficLayer and VirCarlaEnv build and run on Linux, with a 20.04+22.04+24.04 CI gate

yunlishao-vibe:
## Summary

Makes **TrafficLayer** and **VirCarlaEnv** compile, link and *run* on Linux, and adds a CI gate for both on Ubuntu 20.04, 22.04 and 24.04.

The MSBuild/`.vcxproj` build is **untouched** and remains the Windows build of record. Migrating Windows onto CMake is deliberately left as follow-up work — this PR only adds a CMake path beside it, written so that migration is later a deletion rather than a rewrite.

Both targets are verified beyond compiling: SUMO ↔ TrafficLayer ↔ Python echo client, and SUMO ↔ TrafficLayer ↔ VirCarlaEnv ↔ CARLA, both run end to end on Ubuntu 20.04. Windows was re-verified after the TrafficLayer work (Debug+Release plus both DriverModel DLLs); the VirCarlaEnv changes touch no `.vcxproj` and no `ProprietaryFiles`.

### Scope

- **In:** TrafficLayer (SUMO / `libtraci` path) **and** VirCarlaEnv (CARLA C++ client path).
- **VISSIM on Linux:** rejected at config-parse time with an explicit diagnostic rather than left to hang on a `connect()` that can never succeed. The two Windows-only DSProxy TUs drop out of the Linux source list.
- **Out:** the Windows→CMake migration, and any *Carla runtime* in CI — CARLA needs a GPU, so hosted runners get compile + link + `--help` only. The runtime evidence below comes from a dev box.

### The defects were found by compiling, not by grepping

The feasibility audit in #65 enumerated the Win32 surface by search. That method structurally cannot find portability defects that aren't Win32 API calls, and three such defects turned up immediately:

| Defect | Why MSVC accepts it |
|---|---|
| `const double M_PI` member in `TrafficHelper.h` | MSVC leaves `M_PI` undefined without `_USE_MATH_DEFINES`; glibc always defines it, so the declaration becomes a syntax error |
| unscoped enum value `link` in `ConfigHelper.cpp` | collides with POSIX `link(2)`. Made the enum `enum class` so the collision class cannot recur |
| `Timer_t` / `QueryPerformanceCounter` | dead code — declared, frequency queried, never read. Deleted rather than ported; the live timing path uses `steady_clock` |

Sharpest confirmation of the audit's main thesis: `SocketHelper`'s plain-POSIX tier declared `clientAddrLen` as `size_t` where `accept()` needs `socklen_t*` — 64-bit vs 32-bit, so that branch **could never have compiled**. Nobody had ever built it.

### Correctness fixes (second commit)

All pre-existing and affecting Windows equally — none are Linux-only:

- **Descriptor leak on every POSIX socket error path.** `close()` was guarded on `RS_DSPACE` at all 10 sites, so a plain-POSIX build called `shutdown()` and never closed the fd.
- **Every POSIX socket failure logged a bare `"error: "`** with no value, at 18 sites — now reports `strerror(errno)`, so a Linux socket failure is diagnosable instead of anonymous.
- **`TrafficHelper::checkIfEgoExist` returned garbage (UB)** when nothing was subscribed yet — control fell off the end of a non-void function. It gates `SimulationMode: 1` ("wait until the ego enters the network"), so a garbage non-zero could declare the ego present before it exists.
- **`size_t` printed through `%d`** — wrong width out of the varargs.

FIXS code now compiles **warning-free** on gcc 9 (3 remaining warnings are inside vendored yaml-cpp).

### Two design decisions worth reviewing

**`CommonLib/PlatformCompat.h` is header-only, and C++14.** Five `.vcxproj` files compile `ConfigHelper.cpp`, two of them inside the private `ProprietaryFiles` submodule — a new `.cpp` TU would have forced a companion submodule PR and edits to five project files. An inline header changes no project file anywhere. C++14 because CommonLib is consumed by the CarMaker/dSPACE RT targets, which is *why* `ConfigHelper` hand-rolled these helpers on `shlwapi` originally (#65 Q4); the constraint is kept, not relaxed. The Win32 bodies are the originals moved verbatim, so Windows behaviour is unchanged by construction.

**The TU list now exists twice.** That is the cost of keeping MSBuild authoritative on Windows. `scripts/check_source_lists.py` diffs the `.vcxproj` and CMake lists and CI fails on divergence, so "added a TU on Windows, forgot Linux" — which would fail *nothing* on Windows and silently never compile on Linux — cannot pass review.

### VirCarlaEnv: the source cost was three lines; the dependency was everything

Nine of the eleven TUs were already Win32-free — what the #174 backend-agnostic refactor bought. The whole source change is `DataLogger`'s one `CreateDirectoryA` and two `WSAGetLastError` sites in `mainVirCarla`, all routed through the existing `PlatformCompat.h`. Everything else was the SDK.

**Acquisition.** CARLA publishes no release assets at all, so the C++ client SDK exists only inside a built CARLA source tree. It is now packed and published the same way `libsumo` already is:

```bash
scripts/pack_native_deps.sh --component carla --carla-root ~/carla_0.9.15 --publish
scripts/fetch_native_deps.sh --component carla        # ~3 s, SHA-256 verified
```

30 MB, against the 468 MB the Windows `-Mode source` path copies wholesale: only the headers `g++ -M` says are reachable and the archives that actually link, `--strip-debug`'d (`libcarla_client.a` alone goes 217 MB → 12.6 MB and still links).

**The trap, which cost a published asset.** The first one linked here, and on 22.04 and 24.04, and *failed to link on stock 20.04* — the one distro it exists to serve. This box **is** focal, so the existing distro guard passed, but it has `libstdc++6` 13.1.0 installed (normal on a box that builds CARLA), and UE4's clang-10 compiled the client against GCC 13's headers:

| symbol | introduced | from |
|---|---|---|
| `exception_ptr::_M_addref` / `_M_release` | GCC 12 | `libcarla_client.a` (boost::asio) |
| `std::__throw_bad_array_new_length` | GCC 11 | `librpc.a` |

*"Built on the oldest distro"* and *"links against the oldest distro's libraries"* are different claims, and only the first was being checked. Both archives are rebuilt with focal's stock `g++-9` and the asset republished.

The gate now differs by artifact kind, which is the part worth reviewing:

- a **shared library** records the symbol *version nodes* it needs and the loader demands exactly those — so reading them off the `.so` is the whole answer (applied to `libtracicpp.so`, which passes: max `GLIBCXX_3.4.22` ≤ `3.4.28`).
- a **static archive** records unversioned names, so the version is whatever the final link picks. Probing on the packing box over-reports (`condition_variable::wait` binds to `GLIBCXX_3.4.30` here and an older node on focal — both correct) and misses the failure that matters. So the authoritative check is a link inside a stock `ubuntu:20.04` container, with a cheap name screen as the fallback when no container runtime exists.

**Q1 and Q2 from the issue are answered.** Q1 (libc++ or libstdc++): libstdc++, verified on the artifact (`0` `_ZNSt3__1` symbols), and the packer refuses an archive carrying libc++ symbols. Q2 (does a client-only build need `UE4_ROOT`): no — UE4's clang is the *server* half's requirement; the client is libstdc++ and rebuilds with the distro compiler once `Build/` has the third-party deps. That rebuild recipe is now `doc/Carla_Linux_building.md` Part Four.

### Running it found what building could not

`tests/Sumo/Probes/TrafficLayer_SUMO_Carla/run_sumo_carla_demo.sh` (counterpart of the `.bat`) drives SUMO → TrafficLayer → VirCarlaEnv → CARLA on the SimpleLoop scenario. It is **click-to-run**: with no arguments it starts its own CARLA on port 2100, loads `simple_loop.xodr`, runs the co-simulation and takes everything down. Deliberately *not* the conventional port 2000 — a box that develops Carla usually has a session there, and this must never spawn 40 vehicles into it or kill it at teardown. `--carla-port <listening port>` attaches to a server you started instead: no map load, no shutdown. On a CARLA 0.9.15 server with `simple_loop.xodr` loaded: the bridge connects, **spawns 43 SUMO vehicles as CARLA actors**, poses them for the run, and despawns every one at shutdown (0 actors left afterwards).

The first attempt died in its first second: the committed config publishes on **port 440**, which Windows binds happily and Linux refuses without root — `bind() failed: Permission denied`. The staged config shifts any sub-1024 port up by 4000, in **both** places, since `ApplicationSetup.port` and `CarlaSetup.CarlaClientPort` are the two ends of one socket. CI had been green throughout.

### Shipping it: one release, two platform bundles

A release now carries both platforms under one channel tag, selected by platform on the consumer side, so nobody chooses:

```
fixs-build-<ver>-windows-x86_64.zip    TrafficLayer.exe, VirCarlaEnv.exe, carmaker/, vissim/   ← unchanged
fixs-build-<ver>-linux-x86_64.zip      TrafficLayer, VirCarlaEnv, CommonLib/, Carla/  (4.4 MB)
```

**One Linux binary for every supported Ubuntu**, not one per distro. Measured on the CI artifacts and then confirmed by running them: a focal build requires `GLIBCXX_3.4.26` and runs on 20.04/22.04/24.04; jammy and noble builds require `3.4.29`/`3.4.32` and do not run on focal. So `release.yml` builds the shipped bundle in a **stock `ubuntu:20.04` container**, and the jammy/noble legs in `linux-build.yml` stay a compile gate whose artifacts never ship. `pack_linux_bundle.sh` re-checks the binaries themselves, because a box can *be* focal and still carry a newer libstdc++ — this one does, and its locally built `VirCarlaEnv` requires `GLIBCXX_3.4.30` and would not load on stock focal. The gate catches it.

Two zips rather than one because `Compress-Archive` does not store the Unix execute bit (the repo already re-`chmod +x`es bundled `.sh` files for that reason), so the Linux half is packed on Linux by `zip(1)` — and the packer asserts the bit survived, since that is the entire justification for the split.

**The selector had to be fixed first.** #306 had just eliminated `libsumo-*.zip | first` for the native deps; the build bundle was still picked the same way one level up, so publishing a second bundle would have reintroduced the identical bug. `update_fixs.{sh,ps1}` now prefer the platform-qualified name, fall back to the legacy unqualified one **on Windows only** (every unqualified bundle ever published contains `.exe` files), and treat two candidates as an error. The bash selector is unit-matrixed over nine release layouts, evaluating the function straight out of the shipped script so the test cannot drift from it.

### Notes for reviewers

- SUMO 1.22 declares `cmake_minimum_required(VERSION 3.5)`, so focal's stock cmake 3.16 clears the floor. `libtracicpp` needs only `xerces/proj/gdal/zlib/python3-dev`, and building just that target avoids the gdal/fox/osg chain `libsumocpp` would drag in.
- `RealSimVersion.h` was generated **only** by `generate_version.ps1` (PowerShell), so a Linux build had no version header at all. `cmake/FixsVersion.cmake` mirrors its semantics — including the `v[0-9]*` tag match that ignores the rolling `latest`/`alpha_*` tags (#191) — and generates into the build tree, never the source tree.
- `yaml-cpp/dll.h` is a generated artifact (`generate_export_header`), not in git. That's why anything including `yaml.h` fails until yaml-cpp is configured; the CMake target depends on it properly.
- **The PowerShell selector is unverified locally** — this box has no `pwsh`. It mirrors the bash one line for line and needs a Windows run before release.
- **VirCarlaEnv is optional, and a CI skip is a failure.** Configure skips the target when the SDK is absent, so a machine without CARLA still builds TrafficLayer — but in CI the fetch step is supposed to have provided it, so a skip there fails the job rather than quietly shrinking its scope.
- The bridge prints a CARLA version-mismatch warning (`client 0.9.15-dev-dirty / server 0.9.15`). Harmless — the stamp comes from the modified CARLA tree the SDK was built from — but it persists until the asset is built from a clean checkout.
- The pinned SUMO 1.22 **server** does not build on this box (SUMO picks up system libfmt 6 and fails to link on `fmt::v6::format_error`). The co-sim ran against the distro's sumo 1.20 with the 1.22 `libtraci` client, which TraCI accepted. Worth pinning down before anyone depends on `--with-server` here.
- 20.04, 22.04 and 24.04 all run as `container:` images rather than `runs-on:` — `ubuntu-20.04` was retired from hosted runners in 2025 and is only reachable as a container, and pinning all three the same way keeps glibc/gcc determined by the image. 20.04 is the artifact-worthy one: glibc is forward-incompatible only, so a focal binary runs on jammy, not the reverse.

## Related Issues / Tasks

Relates to #65. Both targets the issue scoped are now done; the Windows→CMake migration remains, as does #300 (the 104 Windows libsumo DLLs).

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Maintenance / Refactor
- [x] Documentation

- `CommonLib/` — `PlatformCompat.h` (new), `ConfigHelper.{cpp,h}`, `SocketHelper.{cpp,h}`, `TrafficHelper.{cpp,h}`
- `TrafficLayer/TrafficLayer/mainTrafficLayer.cpp`
- `CMakeLists.txt`, `cmake/FixsVersion.cmake` (new)
- `scripts/check_source_lists.py` (new)
- `.github/workflows/linux-build.yml` (new)

No `.vcxproj` file was modified. `ProprietaryFiles` is untouched — no companion submodule PR is needed.

## Test Cases

Linux (Ubuntu 20.04, WSL2) — every CI step rehearsed locally before the workflow was committed:

```bash
python3 scripts/check_source_lists.py          # OK: 8 translation units agree
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DFIXS_SUMO_HOME=<sumo>
cmake --build build -j                          # 0 warnings in FIXS code
./build/TrafficLayer --help                     # exit 0
./build/TrafficLayer -f tests/Python/SimpleEchoClient/config.yaml
#   -> "Read configuration file success" / "Traffic Simulator: SUMO"
./build/TrafficLayer -f <vissim.yaml>           # exit != 0, "VISSIM is Windows-only"
```

Windows regression:

```
scripts\dispatch\2_core_components.bat    # TrafficLayer Debug + Release: success
scripts\dispatch\3_vissim_components.bat  # both DriverModel DLLs, Debug + Release: success
TrafficLayer\x64\Release\TrafficLayer.exe --help   # v0.9.0
```

Not covered: a live SUMO co-sim on Linux (needs a SUMO runtime configured via `SumoSetup.RuntimeLibraryPath`), and anything Carla.

## Environment
- Python version: 3.8.10 (Linux CI path), 3.x on Windows for the source-list guard
- MATLAB/Simulink/dSPACE version: n/a
- SUMO version: 1.22.0 (`libtraci`; `ENABLE_LIBSUMO` remains commented out)
- VISSIM version: n/a — VISSIM is explicitly rejected on Linux
- CARLA version: 0.9.15 (C++ client SDK, published as `libcarla-0.9.15-linux-x86_64.zip`; boost 1.81)
- IPG CarMaker version: n/a
- Linux toolchain: Ubuntu 20.04.6 / 22.04 / 24.04 (glibc 2.31/2.35/2.39, gcc 9.4/11/13), cmake 3.16.3 on focal

## Checklist
- [x] Tests pass locally
- [x] Version-specific changes are noted (if any)

## Additional Notes

`doc/BUILD.md` is still not touched — it is Windows-only and orphaned from the RTD TOC (#137), so a Linux section belongs there once #137 settles the TOC. `doc/Carla_Linux_building.md` *is* updated (Part Four), because the CARLA SDK rebuild recipe is what the packer's error messages point at and it cannot live in a commit message.


cjffly:
## Update: VirCarlaEnv now builds and runs on Linux too

Picking up the handoff's "next task". The PR title, body and scope are updated — this comment is what changed since the last review pass, in case the earlier body was already read.

**Six commits, smallest first:**

| Commit | What |
|---|---|
| `9ec5044` | `pack_native_deps.sh --component carla` — publish the CARLA client SDK as a 30 MB asset |
| `df32b81` | `fetch_native_deps.sh --component carla` — prebuilt (3.7 s) or from a local CARLA tree |
| `a76e8d5` | the three Win32 sites in VirCarlaEnv's TUs, through `PlatformCompat.h` |
| `6e8a0f9` | the `VirCarlaEnv` CMake target; `check_source_lists.py` now guards both targets |
| `cc48c2d` | `dispatch.sh --with-carla` |
| `62d26b6` | **the ABI fix** — the asset must link on *stock* focal, and now must prove it |
| `d769183` | CI: VirCarlaEnv on all three distros |
| `e20ca64` | the first Linux SUMO↔CARLA co-simulation, and the port bug it exposed |

**The two findings worth a reviewer's attention** (both detailed in the body):

1. **The first published asset was unusable on Ubuntu 20.04** — the distro it exists to serve. It linked on the packing box and on 22.04/24.04. The box *is* focal but carries `libstdc++6` 13.1.0, so UE4's clang built the client against GCC 13 headers and emitted symbols focal's stock libstdc++ never had. Found by running the CI legs in containers locally before pushing them; nothing on the packing machine could have found it. Fixed by rebuilding `libcarla_client.a` and `librpc.a` with focal's `g++-9`, republishing, and adding a gate that link-tests the payload inside a stock `ubuntu:20.04` container.

2. **`bind() failed: Permission denied`** — the SUMO↔Carla demo config publishes on port 440, which is privileged on Linux. CI had been green the whole time; this appeared in the first second of the first real run. That is the same shape as the `SO_REUSEADDR` and `SA_RESTART` bugs from the earlier round: building proves nothing about running.

**Evidence:**

```
[1/3] starting SUMO on TraCI port 1337 ...
[2/3] starting TrafficLayer (SUMO path) ...
[3/3] running VirCarlaEnv for 20s ...
Spawned Carla actor 65 (vehicle.audi.tt)
RESULT: PASS -- bridge connected, spawned vehicles (43 spawn lines), still running at 20s
```

0 actors left in the world after shutdown; 0 errors in `trafficlayer.log`. All three CI legs (20.04/22.04/24.04) were also run locally in containers before pushing: build → link → `ldd` → `--help`, plus the TU guard.

**Answers two of the issue's open questions.** Q1 (client ABI): libstdc++, verified on the artifact, and the packer refuses an archive carrying libc++ symbols. Q2 (client-only build without `UE4_ROOT`): not needed — UE4's clang is the *server* half's requirement.

**Known-open, none blocking:**
- The pinned SUMO 1.22 *server* does not build on this box (system libfmt 6 vs SUMO's bundled fmt). The co-sim ran against the distro's sumo 1.20 with the 1.22 `libtraci` client, which TraCI accepted.
- The libcarla asset is built from a modified CARLA checkout (`0.9.15-dev-dirty`: the boost 1.80→1.81 `Setup.sh` bump), so the bridge logs a harmless version-mismatch warning. Same dev-box provenance weakness the libsumo asset already carries (handoff item 4b) — the fix for both is a `workflow_dispatch` packer job.
- `4a` (TrafficLayer can't be stopped while waiting for its first client) is unchanged and still deferred.

cjffly:
## Update: one release, two platform bundles

Merged `dev_v0.9.0` first, so this sits on top of #306 rather than colliding with it.

**The Linux bundle now rides the same release as the Windows one.** `release.yml` gained `linux-bundle` (stock `ubuntu:20.04` container → build, pack, upload artifact) and `linux-publish` (runner image, which has `gh` → attach to the channel tag). Both key off `build-publish`'s outputs, so the two bundles cannot land on different tags.

```
fixs-build-<ver>-windows-x86_64.zip    unchanged
fixs-build-<ver>-linux-x86_64.zip      TrafficLayer, VirCarlaEnv, CommonLib/, Carla/   4.4 MB
```

**One Linux binary for every supported Ubuntu.** Measured on the CI artifacts, then confirmed by running them: focal requires `GLIBCXX_3.4.26` and runs on 20.04/22.04/24.04; jammy/noble require `3.4.29`/`3.4.32` and do not run on focal. The shipped bundle is therefore built in a stock focal container; the jammy/noble legs stay a compile gate whose artifacts never ship.

**The selector had to be fixed first.** #306 removed `libsumo-*.zip | first` for native deps; the build bundle was picked the same way one level up, so publishing a second bundle would have reintroduced the identical bug. `update_fixs.{sh,ps1}` now prefer the platform-qualified name, fall back to the legacy unqualified one **on Windows only** (every unqualified bundle ever published holds `.exe` files), and treat two candidates as an error. Nine-case matrix in bash, evaluating the function straight out of the shipped script.

### Three bugs found by reproducing CI locally before pushing

| | |
|---|---|
| `--out-dir dist` is relative, and the packer runs `zip` from inside the staging dir | CI would have failed on its first run; every earlier local run passed an absolute path |
| `. /etc/os-release` defines `VERSION` | clobbered the label → `fixs-build-20.04.6 LTS (Focal Fossa)-linux-x86_64.zip` |
| literal `v0.0.0` fallback for a branch with no channel | CI's first bundle was named after a version that does not exist |

The baseline gate also earned its place immediately: this box's locally built `VirCarlaEnv` requires `GLIBCXX_3.4.30` and would not load on stock focal — refused at pack time, which is exactly why the shipped bundle comes from the container.

### CI status

All green on `f6fe9ce5`: Bundle guard, Linux build (20.04/22.04/24.04), Release (`build-publish` + `Linux bundle (focal)`). The CI-produced bundle was downloaded and verified — 4.5 MB, both binaries, execute bit intact through `upload-artifact`.

### Two things still unverified

- **`Publish Linux bundle` is skipped on feature branches** (it needs a release channel), so the `gh release upload` path has not run yet. It will first execute on a push to `dev_v0.9.0`.
- **The PowerShell selector** mirrors the bash one line for line but this box has no `pwsh`. It needs a Windows run before a release is cut.

yunlishao (approved):

Issue #109, opened 2025-11-19
[MAINTENANCE] Automate Carla libcarla dependencies management in build system


yunlishao:
**Task Summary**  
Automate the management of Carla dependencies by integrating the copy and rename operation of `${CARLA_Root}\PythonAPI\carla\dependencies\` to `${FIXS_Root}\CommonLib\libcarla\` into the build system.

**Why is this needed?**  
Currently, per [doc/CARLAdoc.md:35](doc/CARLAdoc.md#L35), users must manually copy the Carla dependencies directory after building LibCarla:

> Once the build is finished, copy the directory `${CARLA_Root}\PythonAPI\carla\dependencies\` into the `${FIXS_Root}\CommonLib\` and rename the directory as **libcarla**.

This manual step is error-prone and breaks the automated build workflow established by the dispatch system. Automating this will:
- Ensure consistent Carla dependency management across development environments
- Reduce setup friction for new developers
- Align with existing build automation patterns (yaml-cpp, libevent, CarMaker, dSPACE)
- Prevent build failures due to missing or incorrectly placed libcarla dependencies

**Scope of Work**  
The automation should integrate into the existing modular build system:

1. **Configuration**: Add Carla path detection to `dependencies.yaml` (similar to existing tool entries)
   - `carla_root`: Path to Carla installation
   - `carla_version`: Version tracking (e.g., "0.9.15")

2. **Detection Script**: Extend [scripts/dispatch/detect_tool_paths.ps1](scripts/dispatch/detect_tool_paths.ps1) to locate Carla installation
   - Search common installation paths: `C:\CARLA`, `C:\Program Files\CARLA`, user-specified paths
   - Validate `PythonAPI\carla\dependencies\` exists
   - Return Carla root path and version

3. **Build Script**: Create new dispatch script `scripts\dispatch\6_carla_dependencies.ps1`
   - Check if `${CARLA_Root}\PythonAPI\carla\dependencies\` exists
   - Copy directory to `${FIXS_Root}\CommonLib\libcarla\`
   - Handle existing `libcarla` directory (backup/overwrite with user confirmation in manual mode)
   - Log operation to `build.log` and `build_summary.log`
   - Skip gracefully if Carla not detected (not all developers need Carla)

4. **Integration**: Update [dispatch.bat](dispatch.bat) to call the new script after external libraries step
   - Execute after `1_external_libraries.bat` (libcarla is an external dependency)
[...]

Issue #70, opened 2025-10-17
[FEATURE] Enable direct SUMO startup via libsumo


yunlishao:
**Is your feature request related to a problem?**
Currently, TrafficLayer cannot successfully start SUMO using libsumo in-process mode. When `EnableAutoLaunch: true` is set with `ENABLE_LIBSUMO` defined, the call to `libsumo::Simulation::start()` causes TrafficLayer to become unresponsive or exit, preventing client connections from being serviced.

**Describe the solution you'd like**
Implement proper libsumo integration to enable direct in-process SUMO startup:

1. Diagnose why `libsumo::Simulation::start()` is causing TrafficLayer to hang or exit
2. Implement proper initialization sequence for libsumo
3. Ensure TrafficLayer's main event loop can properly:
   - Initialize SUMO via libsumo
   - Service client connections
   - Step the SUMO simulation via `SUMO_TRACI_NAMESPACE::Simulation::step()`
   - Exchange data with clients

**Describe alternatives you've considered**
- **Current workaround**: Use external process launch (`#ifndef ENABLE_LIBSUMO`) which successfully launches sumo-gui as a separate process and connects via TraCI. This works reliably but doesn't provide the performance benefits of in-process integration.
- **libtraci**: Use TCP/IP based libtraci instead of libsumo, but this requires a running SUMO instance and loses the simplicity of in-process calls.

**Environment (if feature is environment-specific)**
- C++ compiler/version: MSVC (Windows), GCC (Linux)
- SUMO version: 1.22.0

**Additional context**
Related to PR #69 which implemented the libtraci/libsumo migration. The external process launch works correctly, but the libsumo direct start needs investigation and fixes.

**Current behavior:**
- Path resolution works correctly (config file paths are resolved relative to config directory)
- External process launch (sumo-gui) works successfully
- libsumo mode causes TrafficLayer to exit or hang during `libsumo::Simulation::start()`

**Current code location:**
- `CommonLib/TrafficHelper.cpp:44-61` - libsumo initialization block
- `CommonLib/TrafficHelper.cpp:62-108` - External process launch (working alternative)
- `CommonLib/TrafficHelper.cpp:903-905` - Simulation stepping logic

**Acceptance criteria:**
- [ ] TrafficLayer successfully initializes SUMO via libsumo
- [ ] TrafficLayer remains responsive and services client connections
- [ ] Simulation data is correctly exchanged with clients
- [ ] Works with the SimpleEchoClient test case
[...]

PR #244, opened 2026-08-01
#238/drop vendored libsumo, fetch native deps from release

### LINKED ISSUE #238: [MAINTENANCE] Drop vendored libsumo from git; always fetch native deps from the release zip
## Context

`CommonLib/libsumo/` is committed to git — 105 binary files, ~430 MB in the working tree. `libcarla`
is already gitignored and acquired at setup time. `scripts/dispatch/fetch_native_deps.ps1` was
written with this change in mind; its header already says it acquires libsumo *"when it's dropped
from git later"*, and its prebuilt path fetches `libsumo-<ver>.zip` **"only if libsumo isn't already
vendored"**. So the machinery exists — this issue is about actually flipping the switch.

Motivation reinforced by #70: the vendored copy was silently incomplete (missing `geos_c.dll` and
`geos.dll`) and nobody noticed, because binaries in git get no verification. A fetched zip is
checksum-verified and produced by one reproducible path.

## Proposed change

1. Remove `CommonLib/libsumo/` from git; add it to `.gitignore` alongside `libcarla`.
2. Make `fetch_native_deps.ps1` acquire libsumo unconditionally rather than only when not vendored.
3. Ensure a fresh clone can get to a build without manual steps (see below).
4. Update `doc/BUILD.md` / `doc/DEVELOPER_GUIDE.md` for the new first-run flow.

## The real gap: no single "initialize a fresh clone" entry point

Asked directly — we do **not** have one. What exists is partial and uncoordinated:

| Script | Purpose | Called by `dispatch.bat`? |
|---|---|---|
| `scripts/dispatch/fetch_native_deps.ps1` | libcarla (+ libsumo later) | **No** |
| `scripts/dispatch/1_external_libraries.bat` | builds yaml-cpp | Yes |
| `git submodule update --init` | ProprietaryFiles | No |
| `scripts/fetch_fixs.ps1` | for **consumers** of a FIXS release, not for a dev checkout | n/a |

So today a new developer must know to run `fetch_native_deps.ps1` by hand; `dispatch.bat` only
builds yaml-cpp. If libsumo stops being in git, that manual step becomes mandatory and the failure
mode for missing it is a confusing build/link error.

Proposal: add `scripts/initialize_fixs.ps1` (name TBD) that is idempotent and does:
submodule init -> fetch native deps (libcarla + libsumo, checksum-verified) -> build yaml-cpp ->
report what was found/skipped. Then have `dispatch.bat` call it, so `dispatch.bat` alone works on a
fresh clone. Note `fetch_fixs.ps1` is a different thing (consumer-side) and should not be conflated.

## Acceptance criteria

- [ ] `CommonLib/libsumo/` no longer tracked; `.gitignore` updated
- [ ] `fetch_native_deps.ps1` fetches libsumo unconditionally, SHA-256 verified
- [ ] A fresh clone reaches a successful `dispatch.bat` build with no manual dependency steps
- [ ] Script is idempotent and safe to re-run
- [ ] Release CI still builds (it already uses prebuilt mode)
- [ ] Docs updated

## Notes / risks

- **Do this only after the release asset is correct.** The published `libsumo-1.22.0.zip` is missing
  `geos_c.dll` + `geos.dll` (#70, fixed in #237). Removing the git copy while the zip is broken would
  leave no working source of the DLLs. Re-pack and re-publish first:
  `pack_native_deps.ps1 -Component sumo -Publish`. #237 adds a guard that refuses to pack an
  unloadable `bin/`.
- Repo history keeps the blobs, so clone size only improves for shallow clones — worth stating so
  expectations are right.
- Offline/air-gapped developers need a documented manual path.

Related: #109 (native-deps rolling release), #70 / #237 (the incomplete vendored copy).

yunlishao-vibe:
## Summary

Drops `CommonLib/libsumo` (134 files, ~430 MB) from git and acquires it — like `libcarla` — as a checksum-verified, version-named zip from the public rolling `fixs-native-deps` release. Adds the "initialize a fresh clone" entry point the repo did not have, so `dispatch.bat` alone works on a fresh checkout.

The motivation is verification, not just repo size. The vendored copy shipped for months missing `geos_c.dll` + `geos.dll` (imported by `gdal.dll`, which `libsumocpp.dll` pulls in), so `libsumocpp.dll` could not load at all. Because TrafficLayer *delay-loads* it, the link succeeded and the failure surfaced only at the first libsumo call, as a Win32 loader exception no `catch` block can see (#70). Binaries in git get no verification; a fetched zip is checksummed and produced by one reproducible path.

### Changes

- **`.gitignore` / `git rm`** — `CommonLib/libsumo` no longer tracked.
- **`fetch_native_deps.ps1`** — restructured into two independent components. **libsumo is required** (`TrafficLayer.vcxproj` links `..\..\CommonLib\libsumo\bin\libsumocpp.lib` directly, so nothing core builds without it); **libcarla stays optional**. Previously the script exited early on the libcarla sentinel and exited 0 when Carla was unconfigured — either path would have silently skipped libsumo. `-Component` selects one; libsumo failure is fatal, libcarla failure only warns.
- **`libsumo_verify.ps1`** (new) — shared load test used by both fetch and pack. It `LoadLibraryEx`s each probe DLL instead of checking that files exist. A presence check would *not* have caught the geos gap: all four probed DLLs were present; only an actual load walks the transitive import graph. Packing an unloadable `bin/` is now refused, and a bad asset is caught at fetch with a message naming the fix.
- **`initialize_fixs.ps1`** (new) — idempotent fresh-clone setup: submodule init → native deps → yaml-cpp, with a per-step summary. `dispatch.bat` calls it as step 1. Distinct from `scripts/fetch_fixs.ps1`, which is consumer-side.
- **yaml-cpp guard hardened** — now tests for `Release/yaml-cpp.lib` rather than the build *directory*. CMake creates that directory at configure time, so an interrupted build previously reported "already built" and failed later as an opaque link error. (Found while testing this PR.)
- **`Invoke-Native` wrapper** — PowerShell 5.1 wraps native stderr in a `NativeCommandError`, so one benign line (CMake's deprecation warning, git progress) aborted initialization under `ErrorActionPreference='Stop'`. (Also found while testing.)
- **`release.yml`** — libsumo gets its own cache and a **hard-failing** fetch step; the libcarla step is now `-Component carla` so its soft-warn cannot mask a libsumo failure.
- **Docs** — `BUILD.md` gains a First-Run Setup section (with the offline path, and the note that this only shrinks *shallow* clones since history keeps the blobs); `DEVELOPER_GUIDE.md`, `ContributionGuide.md`, `README.md`, `CLAUDE.md`, `dependencies.yaml` updated.

## Related Issues / Tasks

Closes #238. Depends on #70 / #237 (see blocker below). Related: #109 (native-deps rolling release).

## ⚠️ Blocker — merge order (this is why the PR is a draft)

The published `libsumo-1.22.0.zip` is **still missing the geos DLLs**. Verified directly: 134 entries, zero geos — it was packed on 2026-07-28 from the then-broken vendored tree. Merging #70 repairs the *git-vendored* copy but **not** the release asset, which changes only when someone runs `pack_native_deps.ps1 -Publish`. Since this PR makes that asset the only source of the SUMO runtime, it must not merge first.

Required sequence:

1. **Merge #70** → vendored tree gains `geos.dll` + `geos_c.dll`
2. **Re-pack + publish** from that corrected tree:
   ```powershell
   powershell -File scripts\dispatch\pack_native_deps.ps1 -Component sumo -Publish
   ```
3. **Merge this PR** → vendored copy removed; clones fetch a working zip

Note the overlap with #70: its geos commit vendors those DLLs *into* git, which this PR removes. Only #70's `TrafficHelper.cpp` fixes (dropping `setOrder()` and `--num-clients`) carry forward; the vendoring half becomes obsolete once this lands.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Maintenance / Refactor
- [x] Documentation

- `scripts/initialize_fixs.ps1` (new), `scripts/dispatch/libsumo_verify.ps1` (new)
- `scripts/dispatch/fetch_native_deps.ps1`, `pack_native_deps.ps1`, `dispatch.bat`, `4c_carla_virenv.ps1`
- `.github/workflows/release.yml`, `.gitignore`, `dependencies.yaml`
- `doc/BUILD.md`, `doc/DEVELOPER_GUIDE.md`, `doc/ContributionGuide.md`, `README.md`, `CLAUDE.md`
- `CommonLib/libsumo/` — untracked (134 deletions)

## Test Cases

Run on a Win11 24H2 workstation with VS 2022:

| Check | Result |
|---|---|
| All changed `.ps1` files parse | pass |
| `fetch_native_deps.ps1 -Component sumo` with libsumo present | skips (idempotent), exit 0 |
| Simulated fresh clone: libsumo moved aside, fetch from live release | downloads, SHA-256 verifies, then **correctly rejects** the incomplete asset with an actionable message |
| `initialize_fixs.ps1` from scratch | submodule OK, native deps OK, yaml-cpp built (`Release/yaml-cpp.lib` + `Debug/yaml-cppd.lib`) |
| `initialize_fixs.ps1` re-run | all three steps short-circuit, exit 0 |
| `2_core_components.bat` against non-vendored libsumo | `TrafficLayer.exe` builds; link line confirms `..\..\CommonLib\libsumo\bin\libsumocpp.lib` |
| `pack_native_deps.ps1 -Component sumo` with geos present | load check 4/4, packs 126.7 MB |
| Packed zip vs. published zip | diff is exactly `+geos.dll`, `+geos_c.dll` (136 vs 134 entries) |

Negative control for the load test: with `geos_c.dll` renamed away, `libsumocpp.dll` fails to load (win32 error 126); restored, all four load. This confirms the guard detects the real defect rather than passing vacuously.

> One `2_core_components.bat` run hit the known transient `CL.exe` crash (`-1073741819` / 0xC0000005, documented in BUILD.md for Win11 24H2). It succeeded on retry and is unrelated to these changes.

## Environment

- Python version: 3.x (`realsim_dev`, used only for the zip-content diff)
- SUMO version: 1.22.0
- Visual Studio: 2022 Community (MSVC 14.43)

## Checklist

- [x] Tests pass locally
- [x] Documentation is updated

## Additional Notes

- **Offline/air-gapped developers**: documented in BUILD.md — drop `libsumo-<ver>.zip` into `CommonLib/` out of band so the sentinel exists; `initialize_fixs.ps1` then skips the download.
- **Bumping the SUMO version** in `dependencies.yaml` now requires publishing a matching asset first, or every clone breaks. Called out in `dependencies.yaml`, BUILD.md and DEVELOPER_GUIDE.md.
- The load-test guard overlaps with what #237 proposed; it is implemented here because this PR is what makes an unverified asset repo-breaking.


yunlishao-vibe:
**Blocker cleared — this is now ready for review.**

The merge-order dependency described above has been worked through:

1. **#70 merged** (as #237) — the vendored tree gained `geos.dll` + `geos_c.dll`.
2. **`libsumo-1.22.0.zip` re-published** from the corrected tree. The asset had *not* been repaired by merging #70: release assets only change on upload, so it was still the 134-entry zip packed on 2026-07-28. It is now 136 entries, `updated=2026-08-06`, sha256 `4c362418…`.
3. **`dev_v0.9.0` merged into this branch** (47 commits), and the two geos DLLs dropped from tracking — they now arrive via the fetched zip like the rest of libsumo.

### Conflict resolution worth a reviewer's attention

#237 independently added its own load probe, `Assert-LibsumoSelfContained`, for the same defect this PR guards against. Rather than carry two implementations, they are reconciled into the shared `libsumo_verify.ps1` — which this PR needs anyway, because the **fetch** side must be able to reject a bad asset once the zip is the only source of the SUMO runtime, not merely refuse to publish one.

The merged version takes the better half of each:

| From | What | Why |
|---|---|---|
| #237 | probe the **staged** tree | byte-for-byte what gets zipped, so a staging bug can't slip past |
| #237 | prepend `bin/` to `PATH` | some bundled DLLs are pulled in transitively; `ALTERED_SEARCH_PATH` only covers the loaded DLL's own directory |
| #238 | `FreeLibrary` after each probe | the packer probes its staging dir then deletes it; #237 leaked every handle, risking a sharing violation on cleanup |
| #238 | 4 probe DLLs, not 2 | the Debug variants (`libsumocppD`, `libtracicppD`) ship in the zip too |

### Post-publish verification

| Check | Result |
|---|---|
| Asset on release | 136 entries, both geos DLLs, `updated=2026-08-06T14:48:53Z` |
| Fresh-clone fetch against the **new** asset | downloads, checksum OK, load check **4/4** — this same test failed against the old asset |
| `TrafficLayer.exe` rebuilt against merged sources + **downloaded** libsumo | builds and links clean |

The fetch test ran with `CommonLib/libsumo` deleted, so the build above links against the downloaded tree — not a local leftover.

yunlishao (approved):

Issue #4, opened 2023-07-29
sumo signal light sync with CM signal light


yunlishao:

yunlishao:
![image](https://github.com/Real-Sim-XIL/FIXS/assets/73845081/0e812f8e-b713-4353-be73-e1136159bd7a)
Traffic Objects in CM do not provide C-Code UAQ to change light

PR #253, opened 2026-08-06
FIXS_Applications#4/import_map: the name you import as wins over the export's own name

yunlishao-vibe:
## Problem

`import_map` already generates a CARLA descriptor for a raw RoadRunner export (no `<name>.json`), but it required the export to *already* be named after the map — `<map>.xodr` + `<map>.fbx`. A RoadRunner export never is: it carries the author's working name.

Reproduced against a real export. `MLK_no_signal.zip` holds `MLK_no_signal_0805.fbx/.xodr/.geojson/.rrdata.xml`:

```
stage_package(name='mlk_no_signal')       -> SystemExit: cannot describe 'mlk_no_signal':
                                             no mlk_no_signal.xodr under .../Import/mlk_no_signal
stage_package(name='MLK_no_signal_0805')  -> generated .../Import/MLK_no_signal_0805.json
```

So the feature only worked if you typed the export's own name — and hand-renaming the files was the only way to get a clean map name. The interactive picker made it worse: it asked for "the name matching `<name>.json` inside the package" on a package that has no `.json`.

## Approach: rename the geometry, don't adopt the export's name

Discover the export instead of assuming it — find the staged `.xodr`, pair it with the `.fbx` (or `_Tile_<x>_<y>.fbx` set) beside it — then rename that geometry to the map name being imported.

Renaming rather than naming the map after the export, for two reasons:

1. **Navmesh.** The cook writes the walker navmesh as `Maps/<name>/Nav/<fbx_stem>.bin` (`Util/BuildTools/Import.py`, `build_binary_for_navigation`), while at runtime `FNavigationMesh::Load` asks for `<MapName>.bin`. Geometry whose stem is not the map name silently costs the map its pedestrian navigation.
2. **Identity.** Every re-export would otherwise cook a *different* CARLA map, orphaning saved run setups, app manifests and catalog entries that name the old one. `catalog.json` already treats the cooked `map_name` as a stable identity it can resolve without downloading the bundle.

Only the `.fbx` (and its `.fbm`) moves:

- the `.xodr` keeps the export's name — CARLA copies it to `<name>.xodr` during the cook (`Import.py`, "Move maps XODR files if any"), so renaming it here would be redundant;
- `.geojson` / `.rrdata.xml` stay untouched, so `Import/<name>/` still shows which export the map came from;
- the descriptor records the original stem as `exported_as` (CARLA passes unknown keys through untouched, so this is free).

```
Import/mlk_no_signal.json                            "exported_as": "MLK_no_signal_0805"
Import/mlk_no_signal/mlk_no_signal.fbx               <- renamed
Import/mlk_no_signal/MLK_no_signal_0805.xodr         <- CARLA renames this on cook
Import/mlk_no_signal/MLK_no_signal_0805.geojson      <- provenance
```

## Resolve or ask, never guess

| situation | behaviour |
|---|---|
| export named apart from the map | rename, no prompt |
| several maps in one package | list them, ask which |
| `.fbx` that don't describe one map | list them, ask which |
| same map staged twice | refuse (would fight over one `/Game/<pkg>/Maps/<name>`) |
| `.xodr` with no geometry / no `.xodr` | refuse |
| single-source **and** tiles | refuse (contradiction) |

A failed generation now removes the half-staged export, which the next attempt would otherwise read as the same map staged twice.

## Reaching that path from the picker

- `_infer_package_name` falls back to the lone `.xodr` stem — what `map_name_in` already does — and skips `roadpainter_decals.json`. The local-pick prompt offers that as the default instead of asking for a `.json` name that does not exist.
- `_select_package` asks zip-or-folder and calls `askdirectory()` for the folder case. `choose_map` has always advertised "select a local .zip / **folder**", but `askopenfilename` cannot select a directory, so a typed path was the only way to hand it an extracted export.

## Testing

`tests/Python/unit/test_import_map_raw_export.py` — 23 tests over staging and descriptor generation (single-source, tiled, rename, `TilesInfo.txt`, packaged-map passthrough, every refusal above, name inference, strict tile parsing). Full unit suite: 57 passed.

Also run end-to-end against the two real raw exports on hand — `MLK_no_signal_0805` and `JC_Newest_Ver_MLK_noped1002_final_debug` — both now produce a correct descriptor under the requested map name.

Not covered: the actual UE4 cook. The descriptor shape is unchanged for maps that already imported cleanly (`atlanta_full` cooked from a generated tiled descriptor of exactly this shape).

Refs ORNL-Real-Sim/FIXS_Applications#4

### DIFF
--- a/Carla/README.md
+++ b/Carla/README.md
@@ -183,10 +183,23 @@ CARLA copies it to `<map>.xodr` during the cook — and `.geojson` / `.rrdata.xm
 stay untouched, so `Import/<name>/` still shows which export the map came from.
 The descriptor records it too, as `exported_as`.
 
-Anything genuinely ambiguous is asked, not guessed: several `.xodr` in one
-package, or `.fbx` files that don't describe a single map. Two copies of the same
-map, an `.xodr` with no geometry, or a package that is both single-source and
-tiled are refused outright.
+**One package holds one map** — a tiled map is still one map. That is assumed of
+a Digital-Twin-Library bundle and of a folder you pick by hand, and it is what
+lets the importer resolve a package without asking anything. A package that
+breaks the rule is reported, not guessed at:
+
+| in the package | result |
+|---|---|
+| `.fbx` named after the `.xodr` | imported |
+| one `.fbx` (or one tile set) named *differently* | imported, with a warning naming what it adopted |
+| several `.xodr` — two maps, or one staged twice | refused |
+| several unrelated `.fbx` (a layer-split export) | refused |
+| `.xodr` with no `.fbx`, or `.fbx` with no `.xodr` | refused |
+| both `<name>.fbx` **and** `<name>_Tile_*.fbx` | refused (a map is one or the other) |
+
+The warning case is deliberate rather than silent: a stem mismatch usually means
+the package was assembled by hand, and the lone `.fbx` staged might be scenery
+rather than the road network — which the cook would happily accept.
 
 An app declares its map in a small text file and points the **generic** importer
 at it, so the URL lives in one place (not hard-coded in wrappers):
--- a/Carla/README.md
+++ b/Carla/README.md
@@ -151,6 +151,43 @@ python Carla/import_map.py --package RP_Ver0529 --package-dir C:/Downloads/RP_Ve
 (`--package-dir` and the prompt both accept either the downloaded `.zip` or an
 already-extracted folder.)
 
+### Raw RoadRunner exports
+
+A published bundle carries its own `<name>.json`. A raw export straight out of
+RoadRunner does not — it is just `.fbx` + `.xodr` (+ `.fbm`, `.rrdata.xml`,
+`.geojson`) — so `import_map` writes the descriptor for you: it stages the export
+under `Import/<name>/`, pairs each `.xodr` with its `<x>.fbx` or its
+`<x>_Tile_<i>_<j>.fbx` set, and emits `Import/<name>.json` with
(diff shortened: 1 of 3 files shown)


yunlishao-vibe:
Simplified after review: **one package holds one map** is now stated as an invariant rather than worked around.

The first cut handled a package describing several maps, or several unrelated geometries, by listing candidates and prompting. That was designed for a situation that doesn't exist — a DT-Library bundle hosts one map, a folder picked with `L` is one map, and a tiled map is still one map. It also couldn't work where it mattered: `--import-map`'s paths mostly have no answerable stdin, so both prompts only ever reached their own fallback exit.

| in the package | before | now |
|---|---|---|
| several `.xodr` (two maps, or one staged twice) | prompt → exit | refused, both listed |
| several unrelated `.fbx` (layer-split export) | prompt → exit | refused, candidates listed |
| one `.fbx` / tile set whose stem ≠ the `.xodr` | adopted, quiet `note:` | adopted, **warning** naming what it took |

The warning case stays permissive because one-map-per-package identifies the geometry unambiguously — but it's a warning rather than a note because a stem mismatch usually means the package was assembled by hand, and the lone `.fbx` might be scenery rather than the road network, which the cook would accept without complaint.

Also drops `_pick_one` entirely, and the vestigial "does a stem happen to equal `map_name`" tiebreaker in `_find_export` — a leftover of the very assumption this change removes.

61 tests pass (4 new, covering the warning and both refusals). Both real raw exports still import correctly under the requested name.

yunlishao (approved):

yunlishao (approved):

Issue #258, opened 2026-08-06
[FEATURE] Packaged CARLA: install the DT-Library precooked map instead of requiring a source build


yunlishao:
**Is your feature request related to a problem?**

A user with a **packaged** CARLA (a released `CarlaUE4.exe` / `CarlaUE4.sh` build) cannot run any co-sim on a Digital-Twin-Library map. `run_cosim`'s map preflight only installs a map into a **source** build, so on a packaged build the CARLA slot is silently never filled — the SUMO slot is, the run proceeds, and `load_world` fails on a map that is not there. On Windows the situation is worse: `carla_env_setup.py` does not even offer the packaged option.

This is now an artificial limit. The DT Library **already publishes precooked assets** that a packaged build can consume directly:

```
$ gh release view mlk       --repo ORNL-Real-Sim/Digital-Twin-Library --json assets
  mlk_no_signal.zip                19,238,193
  mlk_no_signal_cooked.tar.gz      29,269,053
$ gh release view roosevelt  --repo ORNL-Real-Sim/Digital-Twin-Library --json assets
  roosevelt_full.zip              379,680,325
  roosevelt_full_cooked.tar.gz    398,748,337
$ gh release view atlanta    --repo ORNL-Real-Sim/Digital-Twin-Library --json assets
  atlanta_full.zip                148,199,893          # no cooked asset yet
```

Verified layout of `mlk_no_signal_cooked.tar.gz` (307 entries) — CARLA's standalone-package layout, and it already follows the conventional path `resolve_cooked_map()` expects:

```
CarlaUE4/Content/mlk_no_signal/Config/mlk_no_signal.Package.json
CarlaUE4/Content/mlk_no_signal/Maps/mlk_no_signal/...      ->  /Game/mlk_no_signal/Maps/mlk_no_signal/mlk_no_signal
CarlaUE4/Content/mlk_no_signal/Static/{Road,RoadLine,Terrain}/...
```

Nothing in FIXS reads these assets. Neither `catalog.json` nor any code path names them.

---

**Where the packaged path is blocked today** (paths relative to repo root, `main` @ 37cb2037)

1. `Carla/run_cosim.py:1938` — the whole import/cook preflight is gated:
   `if not args.no_launch and cfg is not None and cfg.get("mode") == "source":`
   A packaged run skips it entirely. The SUMO slot below (`:2010`) still runs, so the run looks healthy right up to `load_world`.
[...]

Issue #27, opened 2025-07-02
[FEATURE] Carla Lights Placement


kadavparth:
This is the initial implementation for extracting the traffic light information from SUMO and using that information to place signal heads in CARLA using UE4.

Still in development and an initial effort in making the SUMO <-> CARLA light placement pipeline

yunlishao:
Superseded and resolved. The Carla traffic-light placement capability shipped on both `main` and `dev_v0.9.0` via the Carla tooling work — not through PR #53, which was an early snapshot and was closed unmerged.

Current implementation:
- `Carla/utils/extract_sumo_tls_as_table.py` — generalized SUMO `.net.xml` → traffic-light table (any net; adds turn-direction head ordering and far-side-of-junction placement over the #53 prototype)
- `Carla/utils/trafficlight_helper.py` — placement helper (headless/Linux-safe via a carla-import shim)
- `Carla/sumo/auto_place_tls.py`, `Carla/sumo/unreal_placing_tls.py` — Unreal spawning
- `Carla/place_tls.py`, `Carla/sumo/sumo_carla_tl_sync.py` — orchestration

The #53 prototypes (`gen_tl_pose.py`, `sumo2carla.py`, `spawn_ue.py`, hardcoded to Roosevelt.net.xml) were intentionally generalized and dropped. Closing as done.

PR #29, opened 2025-07-03
#26/features/documentation

sarayureddyyy:
## Summary  
Describe what this PR does.  
Documentation for setup environment guide and Test scenario guide

## Related Issues / Tasks  
Closes #26 for documentation update.  
List IDs or links here if applicable.

## Type of Change  
- [ ] Documentation  

## Affected Modules / Components  
SetupGuide.md
ConfigSetup.md 

## Environment  
Only fill in the tools used for testing this PR:  
- Python version:
- MATLAB/Simulink/dSPACE version:
- SUMO version:
- VISSIM version:
- IPG CarMaker version:

## Checklist  
- [ ] Code compiles/runs as expected  
- [ ] Tests pass locally  
- [ ] Documentation is updated (if applicable)  
- [ ] Relevant issues are linked  
- [ ] Version-specific changes are noted (if any)

PR #260, opened 2026-08-06
#258/packaged CARLA installs the precooked DT-Library map

yunlishao-vibe:
## Summary

Lets a **packaged** CARLA run Digital-Twin-Library maps, by installing the precooked asset the library already publishes instead of requiring a source build.

**Mechanism.** A packaged CARLA cannot cook anything — cooking runs the Unreal editor, which a packaged build does not ship. But the DT Library already publishes the *output* of that cook (`mlk_no_signal_cooked.tar.gz`, `roosevelt_full_cooked.tar.gz`), and installing one is a plain extract into the package root. CARLA's own `Util/ImportAssets.sh` is, in full:

```bash
for filepath in `find Import/ -type f -name "*.tar.gz"`; do
  tar --keep-newer-files -xvf ${filepath}
done
```

No manifest, no commandlet, no registration step. **That is the whole basis of this PR**: reimplementing it with `tarfile` is behaviourally equivalent, and it is *necessary* rather than merely tidy — the script is bash-only and CARLA ships no `.bat` counterpart, so shelling out to it would leave Windows with no packaged path at all. This is exactly the limitation `carla_env_setup.py:369` documents and works around by hiding the packaged option on Windows.

Once the map is on disk, nothing else needed new rules. The precooked archive uses the conventional layout (`Content/<name>/Maps/<name>/<name>.umap` plus the same `Config/<name>.Package.json` a cook writes), so `resolve_cooked_map`'s resolve-never-assume logic works unchanged. The only thing that differed between flavours was one path prefix — source keeps Content under `Unreal/CarlaUE4/`, packaged has it beside the launcher — now isolated in `content_root(carla_root, mode)`.

Three changes follow from that:

1. **`import_map.py`** — `content_root()` + a `mode` argument (defaulting from `~/.fixs/carla.json`) on every resolver, so existing source call sites are untouched. `package_root()` resolves through `env.packaged_exe`, so the two accepted spellings of `carla_root` (the package root, or a wrapper holding `WindowsNoEditor/`) cannot produce two different Content dirs. `install_cooked()` extracts with traversal-checked members. `_resolve_carla` now rejects packaged only for operations that genuinely need the editor — cooking and installing were conflated, which is what blocked this outright.
2. **`run_cosim.py`** — a packaged preflight branch beside the source one. It resolves the map's cooked asset, downloads, installs, and resolves the level the same way. Before this, a packaged run skipped the CARLA slot *silently* while the SUMO slot still filled, so the run looked healthy until `load_world`.
3. **`--keep-newer-files` is deliberately not copied.** It exists because the script re-extracts every tarball in `Import/` on each run and must not clobber newer local edits. We install one named asset on purpose; a silent mtime-based skip would leave a half-map.

## Related Issues / Tasks

Refs #258 — **not** `Closes`. Three items in the ticket are deliberately out of scope; see below.

## Type of Change

- [x] New feature

- `Carla/import_map.py` — flavour-aware path resolution, `install_cooked`, cooked-asset naming, `_download_release_asset` (shared by the zip and tar.gz downloaders)
- `Carla/run_cosim.py` — packaged preflight branch; source branch now states `mode="source"` explicitly
- `tests/Python/unit/test_import_map_packaged.py` — new

### 1. Source-build regression — the check that matters most

Every resolver, run from `main` and from this branch against the real source CARLA at `C:/src_ext/Carla`, over all 6 cooked maps present (`RP_Ver0529`, `UGA_v4`, `atlanta_full`, `mlk_no_signal`, `mlk_untextured`, `roosevelt_full`) plus an absent-map case. Compared: `list_imported_maps`, `cooked_content_dir`, `cooked_map_path`, `map_is_imported`, `package_descriptor`, `declared_maps`, `find_level_paths`, `resolve_cooked_map`, `duplicate_level_note`.

```
maps found under C:/src_ext/Carla: ['RP_Ver0529', 'UGA_v4', 'atlanta_full',
                                    'mlk_no_signal', 'mlk_untextured', 'roosevelt_full']
============================================================
IDENTICAL
```

### 2. Precooked install, against the real published archive

`tests/Python/unit/test_import_map_packaged.py`, run against `mlk_no_signal_cooked.tar.gz` downloaded from the DT Library. The packaged CARLA is stubbed (a directory plus a launcher file) — the install only touches the filesystem, so that is faithful — but the **archive is real**, which is the part that could be wrong.

```
$ FIXS_COOKED_TAR=.../mlk_no_signal_cooked.tar.gz pytest tests/Python/unit/test_import_map_packaged.py -q
...... 6 passed
```

Covers: one-prefix difference between flavours; both spellings of a packaged `carla_root` giving the same Content dir; installed package resolving to `("mlk_no_signal", "/Game/mlk_no_signal/Maps/mlk_no_signal/mlk_no_signal")` and agreeing with its own shipped descriptor; re-install a no-op, `force` wiping rather than merging; a tar member escaping the package root refused; cooked-asset convention and catalog override.

Full unit suite: `34 passed, 6 skipped` (the 6 skip without `FIXS_COOKED_TAR`).

### 3. Asset resolution against the live Digital-Twin-Library

Confirms the derived-name fallback matches what is actually published, and that the "no precooked build" path fires for the one map that has none:

```
roosevelt  map=roosevelt_full   derived=roosevelt_full_cooked.tar.gz   published=YES
atlanta    map=atlanta_full     derived=atlanta_full_cooked.tar.gz     published=NO
mlk        map=mlk_no_signal    derived=mlk_no_signal_cooked.tar.gz    published=YES
```

`atlanta` therefore exits before downloading anything, naming the expected asset and listing what the release does publish.

## Environment

- Python version: 3.10 (`fixs_applications` env)
- CARLA version: 0.9.15 source build (`C:/src_ext/Carla`)
- OS: Windows 11

## Checklist

- [x] Tests pass locally
- [ ] Documentation is updated — the user-facing half lives in FIXS_Applications (see below)

## Verified on Linux against a real packaged CARLA

The Limits section of this PR used to say the packaged flow had never been executed. It has now been run end to end on Linux against a packaged **CARLA 0.9.15** (`/home/cave/CARLA_0.9.15_Package`, `VERSION` = 0.9.15), python 3.10 / `realsim`.

### 1. The decisive comparison — `ImportAssets.sh` vs `install_cooked` → **identical**

Two clean copies of the packaged root, the same real `mlk_no_signal_cooked.tar.gz` (29 MB) installed into each — one by dropping it in `Import/` and running CARLA's own `ImportAssets.sh`, one by `import_map.install_cooked`:

```
$ diff -r carla_ref/CarlaUE4/Content/mlk_no_signal carla_ours/CarlaUE4/Content/mlk_no_signal
IDENTICAL (mlk_no_signal)
$ diff -r carla_ref/CarlaUE4/Content carla_ours/CarlaUE4/Content
IDENTICAL (whole Content)

290 files each, 114,911,938 bytes each
```

Content equality is not the whole claim, so metadata was compared too — `tar` applies the umask to archived modes for a non-root user while `tarfile` chmods exactly, which could have diverged. It does not:

```
$ diff <(find carla_ref/... -printf '%y %m %s %T@ %P\n' | sort) \
       <(find carla_ours/... -printf '%y %m %s %T@ %P\n' | sort)
METADATA IDENTICAL      # type, mode, size, mtime, relative path
```

**One real behavioural difference did surface, and it is not visible in that diff.** The archive is not confined to `Content/mlk_no_signal` — it also ships two files under `Content/FIXS/`:

```
-rw-rw-r-- CarlaUE4/Content/FIXS/Props/FIXS_TrafficLight_Head.uasset
-rw-rw-r-- CarlaUE4/Content/FIXS/Props/FIXS_TrafficLight_Head.uexp
```

`ImportAssets.sh` **skipped** both, because `--keep-newer-files` saw newer local copies:

```
tar: Current 'CarlaUE4/Content/FIXS/Props/FIXS_TrafficLight_Head.uexp' is newer or same age
tar: Current 'CarlaUE4/Content/FIXS/Props/FIXS_TrafficLight_Head.uasset' is newer or same age
```

`install_cooked` overwrote both — which is exactly what dropping the flag was meant to do. The trees still matched here only because the archived and local copies are byte-identical (same md5). They need not be in general: **on a package whose `Content/FIXS/` props are newer *and different*, the two paths produce different results**, and `force` does not cover it either — it wipes `Content/<name>`, not the shared `Content/FIXS/`. That is a deliberate, documented consequence of not copying `--keep-newer-files` (an install that silently skipped files would be a half-map), but it is now a measured fact rather than an assumption, and it is worth knowing that these archives reach outside their own map directory.

### 2. End-to-end `run_cosim` — **passes**

`./run_cosim.sh --setup` (chose PACKAGED) wrote `{"mode": "packaged", "carla_root": ...}`, then:

```
[cosim] map 'mlk' -> 'mlk_no_signal'  (DT release mlk)
[import] downloading release 'mlk' from ORNL-Real-Sim/Digital-Twin-Library
[import] installing precooked 'mlk_no_signal' -> .../carla_ours
[import] done: 'mlk_no_signal' installed -> .../Content/mlk_no_signal/Maps/mlk_no_signal/mlk_no_signal.umap
[cosim] note: packaged CARLA - traffic lights and signs cannot be placed here ...
[CARLA] launching (packaged): .../CarlaUE4.sh -carla-rpc-port=2000 -RenderOffScreen
[CARLA] loading world: /Game/mlk_no_signal/Maps/mlk_no_signal/mlk_no_signal
[CARLA] map ready: mlk_no_signal/Maps/mlk_no_signal/mlk_no_signal
```

The co-sim then ran: SUMO TraCI up, vehicles spawning, `--tls-manager sumo` driving the map's lights.

| check | result |
|---|---|
| installs once, `load_world` opens `/Game/mlk_no_signal/Maps/mlk_no_signal/mlk_no_signal` | ✅ |
| second run re-downloads or re-extracts? | ✅ neither — zero `[import]` lines; installed tree and cached tarball byte-for-byte unchanged |
| `--reimport` does | ✅ re-downloads, `removing existing content for a clean re-install`, re-installs; resulting tree still identical to the `ImportAssets.sh` reference |

### 3. Negative path — **passes**, before any download

```
$ ./run_cosim.sh --carla-map atlanta
[cosim] map 'atlanta' -> 'atlanta_full'  (DT release atlanta)
[cosim] no precooked package published for 'atlanta_full' (release 'atlanta' of ORNL-Real-Sim/Digital-Twin-Library).
        A PACKAGED CARLA can only run maps that ship one.
        expected asset: atlanta_full_cooked.tar.gz
        published:      atlanta_full.zip
        Use a source build (run_cosim --setup) to cook this map yourself, or ask for a precooked build of it.
        (exit 1)
```

### 4. Unit tests on Linux — **6 passed / 40 passed**

Exercises `env.packaged_exe`'s Linux branch and the `LinuxNoEditor/` spelling:

```
$ FIXS_COOKED_TAR=.../mlk_no_signal_cooked.tar.gz pytest tests/Python/unit/test_import_map_packaged.py -q
6 passed

$ pytest tests/Python/unit -q                                  # no FIXS_COOKED_TAR
34 passed, 6 skipped
$ FIXS_COOKED_TAR=... pytest tests/Python/unit -q
40 passed
```

### 5. Are traffic lights actually in the published asset? — **yes, 185 of them**

This was flagged as guessed, because the asset is named `no_signal` and placement is impossible on a packaged build. Queried against the running server with the map loaded:

```
map: mlk_no_signal/Maps/mlk_no_signal/mlk_no_signal
actors total: 319
traffic_light actors: 185
speed_limit signs: 0   stop: 0   yield: 0
group-size histogram (heads per intersection): {12: 24, 16: 48, 17: 17, 18: 18, 19: 38, 20: 40}
```

and they are live, not decorative — under `--tls-manager sumo` their states track SUMO:

```
t0: {'Green': 104, 'Red': 75, 'Yellow': 6}
t1: {'Green': 104, 'Red': 66, 'Yellow': 15}
states changed over 10 s: 74/185
```

So **`no_signal` does not mean "no traffic lights"** — the level carries a full set of TL actors (that is also why the archive ships `FIXS_TrafficLight_Head`). What it does lack is OpenDRIVE signals: `world.get_map().to_opendrive()` contains **0 `<signal>` elements**, and there are no sign actors of any kind. TL sync on a packaged build therefore works for this asset; sign-dependent behaviour does not. The `[cosim] note:` line remains correct and remains worth printing — what an asset was cooked with still decides the run — but for `mlk` the answer is now known rather than assumed.

### Not covered

- Only `mlk_no_signal` was installed end to end. `roosevelt_full` (399 MB) resolves through the same code path but was not downloaded and run here.
- The `--setup` folder picker was exercised through its typed-path fallback (`$DISPLAY` unset), not the tkinter dialog.
- Windows itself is unchanged by this run: the Windows packaged option is now offered on the strength of the equivalence in §1, not of a Windows execution.

## Deliberately out of scope

- **SUMO delivery.** The cooked tarball is CARLA-only, so the scenario still comes from the source bundle — roosevelt costs 399 MB cooked + 380 MB bundle for a few hundred KB of SUMO files. Fixing it is a release-shape decision (cooked tarball carries `sumo/`, or a small `<map>_sumo.zip`), tracked in ORNL-Real-Sim/FIXS_Applications#27. The code is commented at the point where the cheaper source would be preferred.
- **The `cooked_asset` catalog field.** Owned by the Digital-Twin-Library repo, specified in FIXS_Applications#27. Read when present; until then derived as `<stem>_cooked.tar.gz`, which §3 above shows matches every published asset today. An explicit empty value means "this map has none" and does *not* fall back to the convention.

## Now included — `carla_env_setup.py` offers packaged on Windows

Previously held back until the equivalence in §1 was proven. It has been, so the gate is gone: `run_setup` offers packaged on every OS, and `--allow-packaged-windows` is kept accepted but hidden from `--help` so anything already passing it keeps working. `Carla/README.md` swaps its Windows-only caveat for the limit that is actually true on both OSes — a package has no editor, so it runs only what the library publishes precooked, exactly as cooked.

The **CARLA-wheel selection needs no matching change.** `ensure_carla`'s packaged branch installs `carla==0.9.15` from PyPI, and PyPI publishes `carla-0.9.15-cp310-cp310-win_amd64.whl`; `environment.yml` pins python 3.10, so the Windows packaged path resolves a wheel today. `find_source_wheel` is source-only and untouched.

### DIFF
--- a/Carla/README.md
+++ b/Carla/README.md
@@ -105,10 +105,15 @@ Setup asks **packaged** vs **source build**, then opens a native folder picker:
   used automatically when it points at a real engine; otherwise you're prompted.
   Launched through the editor as `UE4Editor <uproject> -game`.
 
-**Windows note:** importing a *custom* map into a packaged CARLA is unsupported by
-CARLA (map ingestion is Linux + Docker only &mdash; there is no `ImportAssets.bat`),
-so on Windows setup offers **source build only** for custom-map apps. Pass
-`--allow-packaged-windows` if you only need stock maps (Town01, ...) from a package.
+**What a packaged build cannot do:** cook or place anything &mdash; that needs the
+Unreal editor, which a package does not ship. So it runs a Digital-Twin-Library map
+only if the library publishes it *precooked* (`<map>_cooked.tar.gz`), exactly as it
+was cooked; traffic lights and signs are whatever the asset already contains.
+`run_cosim` installs that asset for you and says so before it launches, and names the
+missing asset up front for a map that has none. To cook a map yourself, use a source
+build. This applies equally on Windows and Linux &mdash; installing a precooked asset
+is a plain extract, done here in `tarfile` rather than through CARLA's bash-only
+`Util/ImportAssets.sh`.
 
 Setup also resolves the **python env** that runs the co-sim and stores it in the
 config, so the launcher works on any machine no matter what the env is named:
--- a/Carla/README.md
+++ b/Carla/README.md
@@ -111,9 +111,18 @@ only if the library publishes it *precooked* (`<map>_cooked.tar.gz`), exactly as
 was cooked; traffic lights and signs are whatever the asset already contains.
 `run_cosim` installs that asset for you and says so before it launches, and names the
 missing asset up front for a map that has none. To cook a map yourself, use a source
-build. This applies equally on Windows and Linux &mdash; installing a precooked asset
+build. *Installing* a precooked asset works the same on Windows and Linux &mdash; it
 is a plain extract, done here in `tarfile` rather than through CARLA's bash-only
 `Util/ImportAssets.sh`.
+
+**Windows note:** the *assets* are not yet OS-neutral. Every `*_cooked.tar.gz` the
+library publishes today is a Linux cook &mdash; its materials carry SPIR-V and no
+Direct3D shaders &mdash; and a packaged build has no shader compiler, so on Windows
+those materials fall back to the default one: correct geometry and traffic lights,
+grey road surface. Setup therefore still keeps packaged behind
+`--allow-packaged-windows` on Windows, and `run_cosim` warns before launching when
+the installed map has no shaders for the platform it is about to run on. Stock maps
+(Town01, ...) are unaffected. Tracked in ORNL-Real-Sim/FIXS_Applications#29.
 
 Setup also resolves the **python env** that runs the co-sim and stores it in the
 config, so the launcher works on any machine no matter what the env is named:
--- a/Carla/carla_env_setup.py
+++ b/Carla/carla_env_setup.py
@@ -366,25 +366,43 @@ def _pick_file(title):
 def run_setup(allow_packaged_windows=False):
     """Interactive setup; writes and returns the config.
 
-    `allow_packaged_windows` no longer gates anything - packaged is offered
-    everywhere. Kept so the flag that scripts and docs already pass stays valid."""
+    On Windows the packaged option is still opt-in (--allow-packaged-windows),
+    but for a different reason than it used to be - see below."""
     print("=== CARLA environment setup ===")
-    # Packaged is offered on every OS. It used to be hidden on Windows because a
-    # custom map could only reach a package through CARLA's Util/ImportAssets.sh,
-    # which is bash-only and has no .bat counterpart. import_map.install_cooked
-    # replaces that script with the same plain extract in `tarfile`, so the route
-    # is now OS-independent; the equivalence was checked by installing the same
-    # mlk_no_signal_cooked.tar.gz both ways into two packaged CARLA 0.9.15 roots
-    # and diffing the resulting Content trees (identical, down to modes+mtimes).
+    # The INSTALL route is now OS-independent: a custom map used to reach a package
+    # only through CARLA's Util/ImportAssets.sh (bash-only, no .bat), and
+    # import_map.install_cooked replaces that script with the same plain extract in
+    # `tarfile`. Installing the same mlk_no_signal_cooked.tar.gz both ways into two
+    # packaged CARLA 0.9.15 roots produces identical Content trees, down to modes
+    # and mtimes.
     #
-    # What stays true on both OSes is the *other* packaged limit: no editor, so
-    # nothing can be cooked or placed here. A packaged build runs only maps the
-    # library publishes precooked, exactly as they were cooked - run_cosim says so
-    # before it launches.
-    print("Which CARLA do you want to use?")
-    print("  [1] Packaged CARLA  (a released build with CarlaUE4.exe / CarlaUE4.sh)")
-    print("  [2] Source build    (run through the Unreal editor: UE4Editor -game)")
-    choice = input("Enter 1 or 2: ").strip()
+    # What is NOT OS-independent is the ASSET. UE4 compiles a material's shaders
+    # per shader platform and stores them in the cooked asset, and every
+    # *_cooked.tar.gz the Digital-Twin-Library publishes today is a LINUX cook: its
+    # four material .uexp carry SPIR-V (GLSL.std.450) and no DXBC anywhere in the
+    # archive, where stock Windows CARLA content carries both. A packaged build has
+    # no shader compiler, so on Windows/D3D those materials have no shader map to
+    # use and fall back to the default one - the map loads, the geometry and the
+    # placed traffic lights are right, and the road surface renders as grey
+    # checkerboard.
+    #
+    # So packaged-on-Windows stays behind --allow-packaged-windows until the
+    # library publishes a Windows cook (or -vulkan is confirmed to resolve these
+    # shader maps on the Windows package, which ships SPIR-V for its own content).
+    # Tracked in ORNL-Real-Sim/FIXS_Applications#29.
+    offer_packaged = platform.system() != "Windows" or allow_packaged_windows
+    if offer_packaged:
+        print("Which CARLA do you want to use?")
+        print("  [1] Packaged CARLA  (a released build with CarlaUE4.exe / CarlaUE4.sh)")
+        print("  [2] Source build    (run through the Unreal editor: UE4Editor -game)")
+        choice = input("Enter 1 or 2: ").strip()
+    else:
+        print("On Windows the map library's precooked maps are Linux cooks: their")
+        print("materials carry no Direct3D shaders, so a packaged CARLA would load")
+        print("them with default (grey) materials. Using source build.")
+        print("(Only need stock maps - Town01, ... - from a packaged build? re-run:")
+        print("   carla_env_setup.py --allow-packaged-windows )")
+        choice = "2"
 
     if choice == "1":
         root = _pick_dir("Select your PACKAGED CARLA folder (contains CarlaUE4.exe / .sh)")
@@ -438,7 +456,9 @@ def main():
     ap = argparse.ArgumentParser(description="Configure which CARLA run_cosim.py uses.")
     ap.add_argument("--show", action="store_true", help="print the current config and exit")
     ap.add_argument("--allow-packaged-windows", action="store_true",
-                    help=argparse.SUPPRESS)  # obsolete: packaged is offered on every OS
+                    help="on Windows, also offer packaged CARLA. The library's "
+                         "precooked maps are Linux cooks and render with default "
+                         "materials there; stock maps (Town01, ...) are fine.")
     args = ap.parse_args()
     if args.show:
         cfg = load_config()
--- a/Carla/carla_env_setup.py
+++ b/Carla/carla_env_setup.py
@@ -364,26 +364,27 @@ def _pick_file(title):
 
 
 def run_setup(allow_packaged_windows=False):
-    """Interactive setup; writes and returns the config."""
+    """Interactive setup; writes and returns the config.
+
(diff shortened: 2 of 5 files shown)


yunlishao-vibe:
## Correction: `1cf2bd55` opened Windows too early — walked back in `3a93ca4d`

That commit flipped the setup gate on the grounds that `install_cooked` removes the `ImportAssets.sh` dependency. **That part holds** — the install route really is OS-independent now, and the byte-identical Linux diff proves it. What it missed is that the route leads to an asset that is not.

UE4 bakes a material's shaders **per shader platform** into the cooked asset. Scanning what the library actually publishes:

| marker | `mlk_no_signal_cooked.tar.gz` (290 files) | stock Windows CARLA 0.9.15 |
|---|---|---|
| SPIR-V / Vulkan (`GLSL.std.450`) | 4 files | present (66–69 per material) |
| D3D bytecode (`DXBC`) | **0 files, entire archive** | present (122–140 per material) |

CARLA's own Windows package carries **both** formats. The library's cook carries only Vulkan — it is a Linux cook.

A packaged build has **no shader compiler**, so on Windows/D3D those four materials (`Default`, `Asphalt1_Road`, `Grass1_Ground`, `LaneMarking1_Marking`) have no shader map to use and fall back to the default one. The level loads, the geometry and the 185 placed traffic lights are correct, and the road surface renders grey. Nothing errors — which is why this is worth gating rather than documenting.

### What changed in `3a93ca4d`

- Windows keeps `--allow-packaged-windows`, now recording *this* reason instead of the `ImportAssets.sh` one this PR genuinely fixes. Linux is unaffected and stays open.
- `import_map.shader_platforms_in()` reports which platforms an installed package carries shaders for; `run_cosim` warns before launch when the running platform is not among them, and names `-vulkan` as the possible out. A warning, never a refusal — it pattern-matches container magic rather than parsing UE4 packages. Only small `.uexp` are read (shader maps live in ~80 KB materials, not the 19 MB `.umap`), so a full scan is 0.11 s.
- `Carla/README.md` states the asset limitation next to the install one.

Detector checked both ways: DT cooked `mlk` → `{'vulkan'}`; stock Windows CARLA content → `{'d3d', 'vulkan'}`.

Unit suite: **42 passed**.

### Still open

`CarlaUE4.exe -vulkan` may resolve these shader maps as-is — the Windows package ships SPIR-V for its own content, so the RHI is evidently built in. Unverified, and cheap to settle: install the asset into a Windows packaged CARLA and launch both ways. If it works, this becomes a documented launch flag rather than a second published asset, and the Windows gate can reopen.

Tracked in ORNL-Real-Sim/FIXS_Applications#29, which also carries the per-platform consequence for #27's `cooked_asset` schema — it cannot be a single string.

yunlishao-vibe:
Superseded by #269, which targets `dev_v0.9.0`.

This branch was cut from `main` @ 37cb2037 — 73 commits behind `dev_v0.9.0` on each side, and GitHub had it as CONFLICTING. #269 is the same work rebased onto the current dev tip, with two changes:

- **`carla_env_setup.py` dropped.** The gate flip in 1cf2bd55 was right that `install_cooked` removes the `ImportAssets.sh` dependency, but the route leads to a Linux-cooked asset, so Windows should keep offering source only — which `dev_v0.9.0` already does. Touching that file would also have regressed the `[3] client` mode added since this branch was cut.
- **`client` mode handled** in the packaged preflight, and the shader-platform detection carried over.

All evidence from here — the Linux byte-identical `ImportAssets.sh` comparison, the end-to-end packaged run, the source-build regression, the shader-platform finding — is reproduced in #269 and re-verified against the dev baseline.

Issue #288, opened 2026-08-07
[BUG] Three Tier-1 carla-env tests are red on dev_v0.9.0 — they call run_cosim.maybe_reexec / ensure_runtime, which moved to carla_env_setup in 4d41e1a7


yunlishao:
**Describe the bug**

Three Tier-1 tests in `tests/Sumo/Carla/test_carla_env.py` fail on `dev_v0.9.0`
because they call `run_cosim.maybe_reexec` and `run_cosim.ensure_runtime`, which no
longer exist on that module:

```
FAILED tests/Sumo/Carla/test_carla_env.py::test_maybe_reexec_noop_same_interpreter
FAILED tests/Sumo/Carla/test_carla_env.py::test_ensure_runtime_noop_when_python_valid
FAILED tests/Sumo/Carla/test_carla_env.py::test_ensure_runtime_repairs_missing_python
3 failed, 96 passed, 8 skipped
```
```
AttributeError: module 'run_cosim' has no attribute 'ensure_runtime'
```

**Root cause**

Commit 4d41e1a7 ("carla env: one interpreter for every entry point, and
--update-python to change it") moved both functions out of `run_cosim` into
`carla_env_setup`, where they now live as `carla_env_setup.reexec_under_configured`
(`Carla/carla_env_setup.py:97`, renamed from `maybe_reexec`) and
`carla_env_setup.ensure_runtime` (`Carla/carla_env_setup.py:295`). The test file was
not moved with them — it still reaches through `run_cosim` at lines 161, 162, 163,
173 and 419.

**Why it matters beyond the red count**

These three are exactly the tests that guard the "every entry point runs under one
interpreter" invariant that 4d41e1a7 introduced. While they error on
`AttributeError` they assert nothing, so that invariant currently has no Tier-1
coverage at all — a regression in `reexec_under_configured` or `ensure_runtime`
would land green.

**To Reproduce**

```
python -m pytest tests/Sumo/Carla/test_carla_env.py -q
```
on a clean `dev_v0.9.0` (reproduced at both `ecfc608a` and `df836064`).

**Expected behavior**

All Tier-1 tests pass. The three should call `env.reexec_under_configured` and
`env.ensure_runtime` — `carla_env_setup` is already imported as `env` at line 22 of
the test file.

**Environment**
- OS: Windows 11
- Python version: 3.10 (`realsim` conda env)
- No CARLA server / GPU needed — these are Tier-1

**Additional context**

[...]

PR #289, opened 2026-08-07
#287/delete the Carla entry points nothing invokes

### LINKED ISSUE #287: [MAINTENANCE] Carla/: delete the 10 .bat/.sh wrappers and the stale sumo/set_spectator_view.py duplicate — nothing invokes them
**Task Summary**

Delete the ten thin `.bat`/`.sh` wrappers under `Carla/` and the superseded
`Carla/sumo/set_spectator_view.py` duplicate, then repoint `Carla/README.md` and the
in-code comments that name them at the `.py` entry points that are actually run.

**Why is this needed?**

Two separate problems, both of which make `Carla/` read as if it has more entry
points than it has.

**1. Nothing invokes the wrappers — in this repo or downstream.**

Each of the ten is a single line with no logic:

```bat
@echo off
python "%~dp0run_cosim.py" %*
```

A `git grep` for all ten names across FIXS returns hits only in (a) the wrappers'
own header comments, (b) `Carla/README.md`, and (c) comments *inside* the `.py`
files describing them. No test, no workflow under `.github/`, and no
`scripts/dispatch/*` build script calls any of them. Three independent paths
confirm the real callers go straight to the `.py`:

- **The application repo's front door execs the `.py` directly.** In
  FIXS_Applications, the root `run_cosim.bat` / `run_cosim.sh` resolve an
  interpreter themselves and then run `"%PY%" "%ROOT%FIXS\Carla\%SCRIPT%"`, where
  `SCRIPT` is one of `run_cosim.py`, `carla_env_setup.py`, `import_map.py`. The
  wrapper layer is bypassed entirely.
- **`place_tls` and `place_signs` are imported as modules**, not shelled out to —
  `run_cosim.py:3346` (`import place_tls`) and `run_cosim.py:3396`
  (`import place_signs`), called in-process.
- **FIXS's own self-update deliberately reaches for the app repo's front door, not
  its neighbour.** `run_cosim.py:156` is `os.path.join(APP_ROOT, "run_cosim.bat")`,
  and `APP_ROOT` is defined at `run_cosim.py:51` as two levels up. The docstring
  already states the principle: *"the ONE thing an app repo is guaranteed to expose
  is its documented entry point."*

**2. `Carla/run_cosim.bat` shares a name with the app repo's real front door.**

This makes the wrappers worse than merely redundant. Anyone who `cd`s into
`FIXS\Carla` and types `run_cosim` gets a materially different program: no
first-run bootstrap gate, no `--sumo-gui`, no `--peer` / `--carla-map` translation,
no `--update-fixs`, and — the one that actually bites — no
`FIXS_ENV_NAME=fixs_applications`. The app front door sets that so an app's extra
packages stay out of the engine's env; without it, `carla_env_setup.py:229` finds
no override and falls through to `environment.yml`'s name, `realsim`. Same command
name, silently different conda env.

**3. `Carla/sumo/set_spectator_view.py` is a superseded duplicate.**

`Carla/set_spectator_view.py` is the generalized replacement — its own docstring
says so: *"Generalized from the ANL set_spectator_view.py (was hardcoded to
Roosevelt)."* The old copy anchors the camera on the traffic-light centroid; the
new one takes `--x/--y/--z` or `--follow <role_name>`. The old copy was even swept
up in the `reexec_under_configured` refactor (`Carla/sumo/set_spectator_view.py:20`
alongside `Carla/set_spectator_view.py:24`) — updated mechanically by a pass that
never noticed there were two of them. `Carla/README.md:20` documents **only** the
old one, so anyone following the README gets the Roosevelt-era version.

**Scope of Work**

Delete:
- `Carla/run_cosim.bat`, `Carla/run_cosim.sh`
- `Carla/setup_carla.bat`, `Carla/setup_carla.sh`
- `Carla/import_map.bat`, `Carla/import_map.sh`
- `Carla/place_tls.bat`, `Carla/place_tls.sh`
- `Carla/place_signs.bat`, `Carla/place_signs.sh`
- `Carla/sumo/set_spectator_view.py`

Rewrite `Carla/README.md` to the `python Carla/<script>.py` form:
- layout block lines 26 / 28 / 30 / 32 (the wrapper entries)
- lines 91 / 93 (`Carla\setup_carla.bat` / `Carla/setup_carla.sh`)
- line 250 (`import_map.bat --package-pick ...`)
- line 266 (`place_tls.bat --map-config ...`)
- line 20 — point it at the top-level `set_spectator_view.py`

Fix the stale comments that name the wrappers:
- `Carla/carla_env_setup.py:22`
- `Carla/import_map.py:1955`
- `Carla/place_signs.py:96`
- `Carla/place_tls.py:189`
- `Carla/run_cosim.py:12`

**Explicitly out of scope**

- **No packaging change is needed.** `scripts/dispatch/8_create_zip.ps1:150` already
  strips `__pycache__` recursively at pack time. A `__pycache__` seen in a consumer's
  `FIXS/Carla/` is generated locally on first run, not shipped.
- `Carla/launch_carla.bat`, `Carla/wait_for_rpc.ps1`, `Carla/carla.env.example` and
  `Carla/load_opendrive_world.py` **stay**. Unlike the wrappers these carry real
  logic — they are the no-Unreal-cook OpenDRIVE path used for CarMaker bridge tests
  (`generate_opendrive_world()`), driven by hand. They are undocumented in the
  README layout block, which is a separate gap, not this cleanup.

**Environment (if version-specific)**

Not version-specific — dead files and docs only, no runtime behavior change.
- CARLA version: 0.9.15 (unchanged)
- Python version: 3.10 (unchanged)

**Risks or Dependencies**

Low. The deleted files have no callers anywhere in FIXS or FIXS_Applications, so
nothing can regress at runtime. The one real risk is **a user's muscle memory**: if
anyone has been typing `Carla\setup_carla.bat` by hand, that path disappears. That
is the intended outcome (see problem 2), and the README rewrite gives the
replacement command on the same line, but it is worth calling out in the release
notes for v0.9.0.

Target branch: `dev_v0.9.0`.

**Additional Context**

Found while auditing `FIXS/Carla/` inside a consumer checkout of FIXS_Applications,
asking which shipped files are still live. Related in spirit to #127 (remove
`archive/` and dead placeholder code), but disjoint in files touched.

yunlishao-vibe:
## Summary

Deletes the ten `.bat`/`.sh` wrappers under `Carla/` and the superseded
`Carla/sumo/set_spectator_view.py`, then makes the README layout block the full
inventory instead of a subset. No runtime code path is touched.

### The wrappers had no callers

Each was a single line with no logic:

```bat
@echo off
python "%~dp0run_cosim.py" %*
```

A `git grep` for all ten names across FIXS returns hits only in the wrappers' own
header comments, `Carla/README.md`, and comments *inside* the `.py` files. No test,
no workflow under `.github/`, no `scripts/dispatch/*` build step calls any of them.
Three independent paths show the real callers going straight to the `.py`:

- An application repo's front door (`run_cosim.bat`/`.sh` at *its* root) resolves an
  interpreter itself and then execs `FIXS\Carla\<script>.py` directly.
- `place_tls` and `place_signs` are `import`ed as modules in-process by
  `run_cosim.py` (`:3346`, `:3396`), never spawned.
- `run_cosim`'s own self-update deliberately reaches for
  `APP_ROOT/run_cosim.bat` (`run_cosim.py:163`) — the *app repo's* front door, not
  its neighbour. Its docstring already states the principle: *"the ONE thing an app
  repo is guaranteed to expose is its documented entry point."*

### But being unreachable isn't why they go — sharing a name is

`Carla/run_cosim.bat` was named identically to the app repo's real front door. A
user who `cd`'d into `Carla/` and typed `run_cosim` got a materially different
program: no first-run bootstrap gate, no `--sumo-gui`, no `--peer`/`--carla-map`
translation, no `--update-fixs`, and no `FIXS_ENV_NAME`. Without that last one
`carla_env_setup.py:229` finds no override and falls through to `environment.yml`'s
name, `realsim` — so the same command silently ran the **engine's** conda env
instead of the app's. That is the failure this removes; the redundancy is only what
made it invisible.

### `sumo/set_spectator_view.py` goes because `run_cosim` absorbed it

> **Corrected after merge.** An earlier revision of this description said this file
> was superseded by `Carla/set_spectator_view.py`. That named the wrong supersessor —
> see the [correction comment](#issuecomment-5220927613). The merged tree is unchanged
> and correct; only the reasoning below needed fixing.

The deleted script's whole job was the **auto-anchor**: centroid of
`traffic.traffic_light*` actors, else the first spawn point, else the origin.
`run_cosim` now does that on every run, and does it better:

- `_frame_from_actors` (`run_cosim.py:1200`) issues the identical actor query and
  returns the same centroid, feeding `position_spectator` (`:1255`).
- That absorbed version also fixes the geometry. The old script parked the camera
  *directly above* the centroid at pitch −55, so the look ray fell short and the
  intersection it was aiming at sat off-frame. `position_spectator` offsets back
  along −X by `height / tan(−pitch)` and scales height to the anchor's span rather
  than using a flat 180 m.
- It runs unprompted at `run_cosim.py:3672` with a richer fallback chain (TL *table*
  → placed actors → whole network), steerable via `--no-spectator`,
  `--spectator-all`, `--spectator-junction`.
- `--load-map` is moot inside `run_cosim`, which loads the target map itself.

`Carla/set_spectator_view.py` is **not** a superset of the deleted file and is not
what replaces it. It is the hand-driven tool for the no-cook OpenDRIVE path
(`--x/--y/--z`, or `--follow <role_name>`), and it has no auto-anchor by design: on a
bare `generate_opendrive_world()` road there are no traffic lights to anchor on, and
the road is built around the origin, which its defaults already target. The two
survivors split the job — `run_cosim` frames co-sim runs automatically,
`set_spectator_view.py` serves the hand-launched server.

Separately, the deleted copy had been swept up in the `reexec_under_configured`
refactor — updated mechanically by a pass that never noticed it was unreachable — and
the README layout block listed **only** it, never the top-level file.

### The README block is now the whole inventory

It previously documented 15 of 28 entries. That is what made live files like
`doctor.py`, `peer.py`, `props.py` and `app_catalog.py` read as dead — an incomplete
inventory is the same defect as a stale one, and it is what prompted this audit in
the first place. Added alongside it is the rule the deletion establishes: every
entry point here is run as `python Carla/<script>.py`, because each re-execs itself
under the configured interpreter, so a wrapper adds nothing and can only shadow the
app's front door.

### Deliberately not in scope

- **No packaging change.** `scripts/dispatch/8_create_zip.ps1:150` already strips
  `__pycache__` recursively at pack time; a `__pycache__` in a consumer's
  `FIXS/Carla/` is generated locally on first run, not shipped.
- **`launch_carla.bat`, `wait_for_rpc.ps1`, `carla.env.example` and
  `load_opendrive_world.py` stay.** Unlike the wrappers these carry real logic — the
  no-Unreal-cook OpenDRIVE path (`generate_opendrive_world()`) used for CarMaker
  bridge tests, driven by hand. They were undocumented in the layout block, which
  this PR fixes rather than treating as evidence they were dead.

## Related Issues

Closes #287

Found while auditing `FIXS/Carla/` from a consumer checkout of FIXS_Applications.
Filed separately from this work: #288 (three Tier-1 `test_carla_env.py` tests are
red on `dev_v0.9.0` — they call `run_cosim.maybe_reexec` / `ensure_runtime`, which
moved to `carla_env_setup` in 4d41e1a7). Left untouched here to keep this a pure
deletion.

## Environment

- Python version: 3.10 (`realsim` conda env)
- SUMO version: n/a — no runtime path touched
- CARLA version: n/a — no runtime path touched

## Checklist

- [x] Tests pass locally
- [x] Issue linked above

## Additional Notes

Tier-1 suite is byte-identical before and after — **96 passed, 8 skipped**, plus the
same 3 pre-existing failures on both the untouched base commit and this branch (see
#288). Verified by running `pytest tests/Sumo/Carla tests/Python/unit` against a
clean checkout of the base and against this branch.

Post-merge, `git grep` for the ten deleted names returns nothing outside
`CommonLib/yaml-cpp`, and every surviving `run_cosim.bat`/`.sh` reference
(`run_cosim.py:153/163/166/190`, `scripts/update_fixs.{ps1,sh}`,
`scripts/dispatch/publish_release.ps1:113`) correctly means the application repo's
front door.

One user-visible consequence worth a line in the v0.9.0 release notes: anyone with
muscle memory for `Carla\setup_carla.bat` will find it gone. That is the intent, and
the README gives the replacement (`python Carla/carla_env_setup.py`) on the same
line.


yunlishao (approved):

yunlishao-vibe:
**Correction to the rationale above — the merged code is right, the reason given for one deletion was not.**

This PR (and the commit message in `d0811f4e`) says `Carla/sumo/set_spectator_view.py`
was superseded by `Carla/set_spectator_view.py`. That names the wrong supersessor.
`Carla/set_spectator_view.py` is **not** a superset of the deleted file: it has no
traffic-light auto-anchor and no `--load-map`; it defaults to `--x 0 --y 0 --z 120`,
top-down. Anyone auditing this by opening that file would find the anchoring missing
and reasonably object.

**What actually replaces it is `run_cosim` itself.** The deleted script's whole job was
the auto-anchor — centroid of `traffic.traffic_light*` actors, else the first spawn
point, else the origin. `run_cosim` now does that on every run:

- `_frame_from_actors` (`run_cosim.py:1200`) issues the identical query,
  `world.get_actors().filter("traffic.traffic_light*")`, and returns the same centroid.
- `position_spectator` (`run_cosim.py:1255`) is also the *correct* version of the
  geometry. The deleted script placed the camera directly above the centroid at pitch
  −55, so the look ray fell short and the intersection it was aiming at sat off-frame.
  `position_spectator` offsets back along −X by `height / tan(−pitch)` so the anchor
  lands in the view centre, and scales height to the anchor's span instead of a flat
  180 m.
- It runs unprompted at `run_cosim.py:3672` with a richer fallback chain (TL *table* →
  placed actors → whole network) and `--no-spectator` / `--spectator-all` /
  `--spectator-junction` to steer it.
- `--load-map` is moot inside `run_cosim`, which loads the target map itself.

**The two survivors split the job, which is why nothing is orphaned:** `run_cosim`
frames co-sim runs automatically, and `Carla/set_spectator_view.py` is the hand-driven
tool for the no-cook OpenDRIVE path (`--x/--y/--z`, or `--follow <role_name>`). The
latter has no auto-anchor *by design* — on a bare `generate_opendrive_world()` road
there are no traffic lights to anchor on, and the road is built around the origin,
which its defaults already target.

No code change follows from this. The merged tree is unchanged and correct; only the
stated reason needed fixing.

Issue #279, opened 2026-08-07
[MAINTENANCE] Compiled REALSIM_VERSION_* still follows git tag topology, and regresses on main


yunlishao:
**Task Summary**

Derive the compiled product version (`REALSIM_VERSION_MAJOR/MINOR/PATCH`, `REALSIM_VERSION_STRING`) from a declared version in the repo rather than from whichever version tag happens to be reachable from `HEAD`.

**Why is this needed?**

`scripts/generate_version.ps1` reads the version from git tags. The only version tag on the 0.9.0 line is the **rolling** `v0.9.0-alpha`, which `publish_release.ps1` deletes and recreates on the built commit every time `dev_v0.9.0` publishes. So the tag lives on a `dev_v0.9.0` commit that is only inside `main`'s history between a release merge and the next dev push.

Measured on the current tips (after f24d0c50 on `main` and 5bb44a70 on `dev_v0.9.0`):

```
$ git rev-list -n1 v0.9.0-alpha        -> 5bb44a70   (dev_v0.9.0 only)

$ git tag --merged f24d0c50 --list 'v[0-9]*'   # main
v0.6.0
v0.7.0

$ git tag --merged 5bb44a70 --list 'v[0-9]*'   # dev_v0.9.0
v0.6.0
v0.7.0
v0.9.0-alpha
```

So `main`'s next build stamps `REALSIM_VERSION_STRING "0.7.0"` into binaries that carry 0.9.0 content, and it will flip back to 0.9.0 at the next release merge and away again at the next dev push. The version a binary reports is not a property of that binary — it is a property of when someone last pushed to another branch.

f24d0c50 already changed the selection from *nearest* tag to *highest reachable* tag, which removes the case where a merge brings a lower tag closer. It cannot fix this case: when the tag leaves `main`'s history entirely there is nothing higher than `v0.7.0` left to find.

Release **asset names** are no longer affected — f24d0c50 names them after the channel (`fixs-build-latest.zip`, `fixs-build-v0.9.0-alpha.zip`). This issue is only about the version compiled into the binaries.

**Scope of Work**

- `scripts/generate_version.ps1`
- wherever the declared version lives (new `VERSION` file, or a key in `dependencies.yaml`)
- `scripts/dispatch/release_channel.ps1` — the channel already knows the train's version and could cross-check the declared one

**Proposed approach**

Declare the version explicitly; let git supply only traceability.

[...]

Issue #276, opened 2026-08-07
[FEATURE] import_map --pick-release dead-ends on a packaged CARLA instead of installing the precooked package


yunlishao:
**Is your feature request related to a problem?**

On a machine configured with a packaged CARLA, `run_cosim --import-map` stops with:

```
[import] the configured CARLA is PACKAGED; cooking a custom map from fbx/xodr needs a
         SOURCE build (run setup_carla and pick source).
         A map published with a precooked asset installs here without one - run it
         through run_cosim, which picks that path automatically.
```

The refusal to *cook* is correct — cooking runs the Unreal editor, which a packaged build does not ship. But the message is telling the user that the thing they asked for is possible, just not from the command they used. `import_map.install_cooked` already installs a Digital-Twin-Library `*_cooked.tar.gz` into a packaged build (a plain extract, no editor), and `_resolve_carla(need_source=False)` already exists precisely so that path stays open. It is wired only into `run_cosim.py` (the `cfg.get("mode") == "packaged"` preflight), so the front door named `--import-map` cannot reach it.

**Describe the solution you'd like**

Route `import_map.py`'s picker to the flavour-appropriate path instead of gating the whole CLI on `need_source=True`: for a packaged config, resolve the catalog's cooked asset for the chosen release, check it is published, download it and call `install_cooked` — the same sequence `run_cosim.py` runs. The pieces (`catalog_cooked_asset`, `release_assets`, `download_cooked_tar`, `install_cooked`, `resolve_cooked_map`) all exist and are already used together there; this is mostly about not duplicating that block but sharing it.

The existing errors stay where they are genuinely right: a *source bundle* handed to a packaged build still cannot be cooked, and a map whose release publishes no cooked asset should still say so by name before downloading anything.

**Describe alternatives you've considered**

Leaving it: the current message does name a working path (`run_cosim`), so nothing is unreachable. The cost is that the command whose entire purpose is importing a map refuses to import one on a flavour that can, which reads as a limitation of the product rather than of the command.

**Environment (if feature is environment-specific)**

Packaged CARLA (`~/.fixs/carla.json` `mode: packaged`) — Linux or Windows. Source builds are unaffected.

Found while fixing the env-consistency defects in #275; deliberately left out of that PR to keep it to one thread.

yunlishao:
Landed on `main` in https://github.com/ORNL-Real-Sim/FIXS/commit/2c60d97f via #278.

[...]

Issue #272, opened 2026-08-07
[MAINTENANCE] FIXS should own the update path: move update_fixs.{ps1,sh} in, collapse fetch_fixs.ps1


yunlishao:
**Task Summary**

Make FIXS own the update path. Move `update_fixs.ps1` and `update_fixs.sh` into this repo's `scripts/`, self-contained and symmetric, parameterized by the consuming app's root. `scripts/fetch_fixs.ps1` folds into `update_fixs.ps1` and is deleted -- once the whole updater is engine-owned, the two-layer split has no remaining purpose.

Companion: ORNL-Real-Sim/FIXS_Applications issue for the consumer-side bootstrap.

**Why is this needed?**

Today the update logic is split by platform AND by repo, and the halves have drifted:

```
run_cosim.bat --update-fixs            run_cosim.sh --update-fixs
  -> APP:scripts\update_fixs.ps1         -> APP:scripts/update_fixs.sh
       -> downloads + runs                    -> does the fetch itself
          FIXS:scripts/fetch_fixs.ps1              (self-contained, 248 lines)
          at the release tag
```

`update_fixs.ps1` <-> `update_fixs.sh` is a correct counterpart pair; all three entry points (`run_cosim.bat`, `run_cosim.sh`, `Carla/run_cosim.py` self-update) dispatch at exactly that boundary. The asymmetry is the second layer: `fetch_fixs.ps1` has no `.sh` counterpart, so a change to release packaging can be applied to one path and not the other.

That is what happened. #238 removed `CommonLib/libsumo/bin` from git and moved it to the `fixs-native-deps` release. `update_fixs.sh` was ported to fetch the packed asset over HTTPS; `fetch_fixs.ps1` still sparse-clones the deleted path:

```
git clone --depth 1 --filter=blob:none --sparse --branch <tag> ...
git sparse-checkout set CommonLib/libsumo/bin      # matches nothing
```

The clone succeeds, sparse-checkout yields an empty tree, the DLL count is 0, and the script hard-fails. Verified against the live tag: clone exit 0, `git ls-tree -r v0.9.0-alpha -- CommonLib/libsumo/bin` = 0 files. `run_cosim.bat --update-fixs` is therefore broken on Windows while `run_cosim.sh --update-fixs` works.

[...]

Issue #266, opened 2026-08-07
[BUG] Vehicle transforms are sent with async ApplyBatch, so they can miss the tick that renders them


yunlishao:
**Describe the bug**

`CarlaBackend::flushBatch()` sends every mirrored vehicle's transform with `client_->ApplyBatch(batch_, false)`. On the CARLA client that is `AsyncCall` — the commands are handed to the socket and the call returns without waiting for the server to apply them:

```cpp
// LibCarla/source/carla/client/detail/Client.cpp
void Client::ApplyBatch(std::vector<rpc::Command> commands, bool do_tick_cue) {
    _pimpl->AsyncCall("apply_batch", std::move(commands), do_tick_cue);
}
```

The very next thing the bridge loop does is `world.Tick()`. So it is racing its own message. When the tick wins, that frame renders **every mirrored vehicle at its previous pose**, while the spectator — placed from `lastAppliedPose`, i.e. the pose the bridge just *commanded* — has already moved on. The next tick applies both, so the vehicle jumps twice as far to catch up.

Visually this is the followed vehicle shuddering back and forth against a smooth camera. It is not the 10 Hz stepping of #261; it is on top of it, and it is the part that reads as *flicker* rather than as *choppiness*.

**To Reproduce**

1. Any scenario with `EnableSpectatorFollow: true` and `CarlaSetup.CarlaTimeStep: 0.1` (so every tick carries a batch).
2. Run and watch the followed vehicle, or attach a read-only second client sampling `world.on_tick` and compare the spectator and the followed actor from the same snapshot.
3. The vehicle's per-tick motion alternates 0.000 / 0.582 m while the camera holds a steady 0.291 m.

Measured on `mlk_eco_driving` / `mlk_no_signal`, `tick 0.1`, `RealtimePacing: true`, 449 consecutive rendered frames:

```
sim_t      ego_step  cam_step   offset
308.0424     0.291     0.291    0.000
308.1424     0.000     0.291    0.291   <- vehicle frozen, camera moved on
308.2424     0.582     0.291    0.000   <- vehicle catches up with a double step
308.6424     0.000     0.291    0.291
308.7424     0.582     0.291    0.000
```

| | frames where the vehicle did not move | frames lagging the camera by >= half a step | offset mean / max |
|---|---|---|---|
| `ApplyBatch` (today) | **108/449 = 24.1%** | 130/449 = 29.0% | 0.089 m / 3.211 m |

[...]

PR #291, opened 2026-08-07
#290/close issues when the code reaches main, not when a PR merges

### LINKED ISSUE #70: [FEATURE] Enable direct SUMO startup via libsumo

### LINKED ISSUE #129: [MAINTENANCE] Consolidate duplicate VISSIM driver model code into shared base

### LINKED ISSUE #131: [MAINTENANCE] Build system cleanup: remove unused libevent, consolidate dispatch scripts, fix version tracking
> **⚠️ Rescoped 2026-07-28 — retargeted to `dev_v0.9.0`.** The repo has moved to versioned dev branches (`dev_v0.8.0`, `dev_v0.9.0`); `dev_v0.9.0` is the active integration branch. **#191 (now closed) has already landed a large release/CI expansion on `dev_v0.9.0` that overlaps this issue's dispatch scope**, changing what remains here. See the "Current state on `dev_v0.9.0`" section below before starting.

> **Related: #191** folded in the version-tracking **correctness** fixes — the `RealSimVersion.h` `0.0.0 / UNKNOWN` bug and the semver-tag-shadowing fix (`git describe --tags --match 'v[0-9]*'`) are **already applied** to `scripts/generate_version.ps1` on `dev_v0.9.0`. What #191 did **not** do is *wire that script into the build/CI pipeline* — that remains here. The libevent removal, libsumo integration, and dispatch renumbering also remain unique to this issue.

## Current state on `dev_v0.9.0` (verified 2026-07-28)

#191 added these to `scripts/dispatch/`, expanding the pipeline well beyond the linear 1–8 scheme this issue assumed:
- `4c_carla_virenv.ps1` — Carla libcarla + VirCarlaEnv (public, #109)
- `fetch_native_deps.ps1`, `pack_native_deps.ps1` — native-dep bundle
- `pack_binaries.ps1`, `bundle_key.ps1`, `msg_contract_hash.ps1` — ProprietaryBinaries bundle / guard
- `publish_release.ps1` (moved into `dispatch/`) — `gh release create latest`
- `.github/workflows/release.yml`, `bundle-guard.yml` — automated CI

Consequences: the "renumber 1–8" plan (item 2) is now partly obsolete and higher-risk — several new scripts are un-numbered helpers, not sequential build steps. `generate_version.ps1` is fixed but still **called by nothing**. `libevent` is still fully present.

## Task Summary

The build system has accumulated technical debt: `libevent` is vendored but marked "Not currently used", dispatch scripts have grown organically without consistency, `RealSimVersion.h.in` uses the old branding, and the `generate_version.ps1` script is disconnected from the dispatch pipeline.

## Why is this needed?

- `CommonLib/libevent/` is a large vendored library (50+ headers) noted as "Not currently used" in `dependencies.yaml` — it should be removed from the repo or gated behind a proper optional flag
- Dispatch scripts have inconsistent numbering/naming (`5a`/`5b`/`4c` sub-letters, plus un-numbered `pack_*`/`bundle_*`/`fetch_*` helpers alongside sequential build steps)
- `scripts/generate_version.ps1` exists (and is now correct, per #191) but is not called by `dispatch.bat` or `release.yml` — version generation is still a manual step
- `RealSimVersion.h.in` uses `REALSIM_VERSION_*` prefix (tracked in issue #128 for renaming, but the version generation script also needs updating)
- `build_libsumo.ps1` and `build_sumo_executables.ps1` are in `scripts/` root but not integrated into `dispatch.bat`
- The `dispatch.bat` build log goes to `scripts/dispatch/build.log` but there's no CI artifact collection and it isn't gitignored

### 1. Remove libevent from CommonLib
- Remove `CommonLib/libevent/` directory (50+ vendored headers, never used)
- Remove libevent from `dependencies.yaml`
- Remove libevent build from `scripts/dispatch/1_external_libraries.bat`
- Verify no source files include libevent headers (grep across all .cpp/.h)
- Add note to doc/BUILD.md about what libevent was and why removed

### 2. Rationalize dispatch script naming (rescoped)
- **Do not** do a naive 1–8 renumber — #191 added `4c` and several un-numbered helper scripts. Instead: document the two script *classes* (sequential build steps vs. release/bundle helpers) and make naming consistent within each class without breaking the release pipeline.
- Ensure `dispatch.bat` calls all sequential build-step scripts in correct order
- **Wire `scripts/generate_version.ps1` into the pipeline** (call from `dispatch.bat` before build, and/or from `release.yml`) so `RealSimVersion.h` is regenerated automatically. The script's correctness fixes already landed via #191 — only the wiring is missing.
- Document each script's purpose in `doc/BUILD.md`

### 3. Integrate libsumo build scripts
- Move `scripts/build_libsumo.ps1` and `scripts/build_sumo_executables.ps1` into `scripts/dispatch/` or document them as standalone utilities
- Add them to BUILD.md with clear instructions on when to use them

### 4. Add auto-test — **deferred to #57**
Originally scoped a `tests/Python/test_build_system.py`. Deferred: there is no pytest CI job today (#57 tracks GitHub Actions for the Python suite), so a build-hygiene test would run nowhere and protect nothing. Its key invariant (dispatch → `generate_version` wiring) is already enforced by CI running `dispatch.bat`. Reintroduce a build-hygiene test alongside #57, when a runner actually executes it, covering:
  - No libevent headers included in any source file
  - Dispatch build-step scripts exist and are referenced in dispatch.bat
  - `dependencies.yaml` is valid YAML with required keys
  - `RealSimVersion.h` is (re)generated by the build with the expected macros

### 5. Update .gitignore
- Ensure `build/` output directory is properly gitignored (except prebuilt binaries intentionally committed)
- Add `*.log` for build log files in `scripts/dispatch/`

## Files Involved
- `CommonLib/libevent/` (delete)
- `dependencies.yaml`
- `scripts/dispatch/*.bat`, `*.ps1`
- `scripts/dispatch/dispatch.bat`
- `scripts/generate_version.ps1`
- `scripts/build_libsumo.ps1`, `scripts/build_sumo_executables.ps1`
- `.github/workflows/release.yml` (version-generation wiring)
- `doc/BUILD.md`
- `.gitignore`
- `tests/Python/test_build_system.py` (new)

## Acceptance Criteria
- [ ] `CommonLib/libevent/` removed from repo
- [ ] No source file includes libevent headers
- [ ] Dispatch script naming rationalized (build-steps vs. helpers) and all build-step scripts called from `dispatch.bat` — without breaking the #191 release pipeline
- [ ] `generate_version.ps1` wired into the build/CI pipeline (correctness fixes already in via #191)
- [ ] libsumo build scripts integrated or documented
- [ ] ~~`tests/Python/test_build_system.py` passes~~ — build-hygiene test deferred to #57 (no pytest CI runner yet)
- [ ] `.gitignore` covers `build/` and dispatch `*.log`
- [ ] `doc/BUILD.md` updated to reflect changes

## Branch and PR Target
- **All PRs for this issue target `dev_v0.9.0`** (the active integration branch). ~~Formerly `dev`~~ — the repo now uses versioned dev branches.
- `dev_v0.9.0` already contains #191's release/CI work; base new PRs on current `dev_v0.9.0` HEAD.
- Suggested PR split: (a) libevent removal, (b) `generate_version.ps1` wiring + `.gitignore`, (c) libsumo integration + BUILD.md, (d) `test_build_system.py`.

### LINKED ISSUE #261: [BUG] CarlaSetup.TrafficRefreshRate defaults to 0.1 s when absent, so a sub-stepped world still moves at 10 Hz
**Describe the bug**

`CarlaSetup.TrafficRefreshRate` is the pose re-apply cadence. `VirCarlaEnv/mainVirCarla.cpp` documents its contract as *"Absent/0 -> every tick"*, and `Carla/run_cosim.py` assumes the same. `CommonLib/ConfigHelper.cpp:654-659` does the opposite: when the key is **absent** it sets the value to **0.1**.

The effect is that sub-stepping the CARLA world buys nothing visually. With `CarlaSetup.CarlaTimeStep: 0.025` the bridge announces `interpolated 4x`, but traffic poses are re-applied only every 0.1 s, so every mirrored vehicle holds a bit-identical pose for 3 rendered frames and jumps a whole feed interval of travel on the 4th. The interpolation code in `VirEnvCore.cpp` is reached once per feed instead of once per tick, so it never interpolates anything.

Under `EnableSpectatorFollow` this is what a user reports as flicker: the camera jumps *with* the followed vehicle, so the vehicle holds the centre of frame correctly while the entire road and surrounding traffic lurch at 10 Hz underneath a 48 fps window.

The two banners printed in the same run disagree, which is the quickest tell:

```
[cosim] cadence (cpp): FIXS feed 0.1 s | CARLA tick 0.025 s (interpolated 4x) | pose refresh 0.025 s | ...
Cadence:               FIXS feed 0.1 s | Carla tick 0.025 s (interpolated 4x) | pose refresh 0.1 s   | ...
```

**To Reproduce**

1. Any scenario yaml with `CarlaSetup.CarlaTimeStep: 0.025`, `EnableSpectatorFollow: true`, and **no** `TrafficRefreshRate` key at all (a generated yaml has none — `run_cosim.py` only rewrites the key when it is already present).
2. Run the co-sim; watch the followed vehicle in the CARLA window.
3. VirCarlaEnv's own banner prints `pose refresh 0.1 s` while run_cosim prints `pose refresh 0.025 s`.
4. The followed vehicle stays centred, but the world advances in visible 10 Hz steps.

Measured on `mlk_eco_driving` / `mlk_no_signal`, shipped `v0.9.0-alpha` bundle (072ec889), over 2161 consecutive rendered frames sampled by a read-only second client on `world.on_tick`:

| | absent key (effective 0.1 s) | explicit `TrafficRefreshRate: 0.0` |
|---|---|---|
| frames on which the world advanced | 540 / 2160 (25 %) | 1774 / 1908 (93 %) |
| gap between advances | exactly 4 frames, 539 times of 539 | 1 frame (1639), 2 frames (134) |
| motion per advancing frame | 0.281 m mean | 0.077 m mean |
| camera-to-vehicle offset, max | 0.290 m | 0.073 m |
| render rate / sim rate | 48.0 fps, 1.20× realtime | 42.4 fps, 1.06× realtime |

The follow itself is correct in both: camera height above the followed vehicle measured 50.000 m with a standard deviation of 0.000 m.

**Root cause**

- `CommonLib/ConfigHelper.cpp:654-659` — absent key ⇒ `CarlaSetup.TrafficRefreshRate = 0.1`.
- `VirCarlaEnv/mainVirCarla.cpp` — `poseRefresh = (cs.TrafficRefreshRate > 1e-9) ? cs.TrafficRefreshRate : carlaStep;` with the comment *"Absent/0 -> every tick"*. The 0.1 default wins, so the fallback to `carlaStep` is unreachable for an absent key.
- `CommonLib/VirEnvCore.cpp:299` — `refreshRate = (int)(1.0 / trafficRefreshRate_)` ⇒ 10, and the pose block runs only when `refreshSlot` changes.
- `Carla/run_cosim.py:3111-3130` — clears a stale key and prints *"the default is every CARLA tick"*. That statement is only true because it writes the key explicitly as `0.0`; it does not apply to a yaml that never had the key.

For comparison, the CarMaker counterpart at `ConfigHelper.cpp:380-385` defaults to 0.001. That asymmetry suggests the Carla 0.1 is a leftover from when this key carried the feed period, before #219 redefined it as the re-apply cadence.

**Suggested direction**

Make the absent-key default 0 (i.e. every tick), so `ConfigHelper` agrees with what `mainVirCarla.cpp` documents and what `run_cosim.py` already assumes. Setting the key would then mean only what its current documentation says: re-apply poses *less* often than the tick, as a cost knob on a heavy scene. The measured cost of moving from 10 Hz to per-tick re-apply on this scenario was 48.0 → 42.4 fps.

**Residual, probably a separate defect**

With per-tick refresh, 134 of 1908 frames still did not advance, and they land only on sub-step 3 (84 frames) and sub-step 0 (50 frames) of the 4-sub-step feed cycle — i.e. at the feed seam, where `prev_`/`next_` shift and the interpolation fraction resets. 96 of those are also the frames where the camera sits more than 2 cm off the vehicle. Much smaller than the main effect, but it is a real seam and it may be the same family as #169.

**Environment**

- C++ compiler/version: MSBuild 17.14.51.32402 (Release, x64), shipped bundle v0.9.0-alpha (072ec889)
- CARLA version: 0.9.15
- SUMO version: 1.22.0
- Python version: 3.10

**Additional context**

Related to #254 (spectator placed before the tick that renders it). #254 removed the one-frame camera lag; this issue is the remaining reason the scene still moves in 10 Hz steps, and the two together determine whether a follow-cam view looks smooth.

### LINKED ISSUE #264: [BUG] run_cosim setup menus mark a guessed value as (current) — Enter silently switches the scenario yaml, e edits the wrong file
## What happens

A saved setup names a scenario yaml, but the scenario menu marks a different one as
`(current)`, offers it as what Enter keeps, and edits it under `e`:

```
[cosim] Run setup 'mlk_eco_driving_ship'
   3) scenario  exp_fix_follow_on.yaml    <- what the setup runs

[cosim] Scenario config for MLK arterial eco-driving on 'mlk_no_signal':
   1) config_Sumo_Carla_ecoDriving.yaml   SUMO + CARLA, eco-driving  (current)   <- not it
   ...
   6) exp_fix_follow_on.yaml              variant in this app's folder
[cosim] Which? [1-8 / e], Enter = keep config_Sumo_Carla_ecoDriving.yaml:
```

## Why

`_config_menu` in `Carla/run_cosim.py` takes the current yaml to be the first row of
the list it just built (`current = options[0][0]`), and `edit_config` is never handed
`rec["config"]`, so the menu has no access to what the setup actually holds. The two
screens can only agree by coincidence.

## What it breaks

1. `(current)` and `Enter = keep ...` name the wrong file.
2. `e` opens the wrong file — a user editing "the current one" edits a config they are
   not running.
3. `_config_menu` **returns** that guess and `_edit_slots` writes it into
   `rec["config"] / rec["config_scope"]`. So opening the scenario row and pressing
   Enter at a prompt that says *keep* silently switches the setup's scenario and flips
   its scope to `app`. Nothing on screen says the setup changed.

The same shape is in the other two menus that own a setup slot:

- `app_catalog.choose_app` — no current mark, `Enter = 1`. Opening the app row on a
  setup running any other app and pressing Enter switches the app, and `cascade()`
  then invalidates the map and the scenario with it.
- `import_map.choose_map` — `Enter` takes the app's map, never the map the setup is
  already running.

`engine`, `carla` and `sumo` are not affected: they read live state through
`derived_from_yaml` and pass it into `_menu(..., current)` correctly.

Separately, `run_profile._fmt` renders the scenario row's scope from `config_scope`,
which stores only `app` / `map`. `map` means "not one of the app's staged yamls", but
it was being printed as *"generated, this app on this map"* — so a yaml the user
dropped in by hand was labelled generated, and the summary disagreed with the menu
("variant in this app's folder") about the same file.

## Fix

- `run_cosim.edit_config(..., current=None)` → matched case-insensitively against the
  list (the path round-trips through `run_profiles.json`), then passed to
  `_config_menu(who, options, current)`. Two cases are called out rather than silently
  falling back: a yaml in use from outside the app folder is appended to the list so it
  stays selectable, and a yaml that no longer exists prints a note. When there is no
  current yaml the first row is marked `(default)`, not `(current)`.
- `_edit_slots` passes `rec["config"]`, `rec["app"]`, `rec["map"]` as the current
  values, and drops the ones an earlier slot in the same pass invalidated — mirroring
  `cascade()`, so the previous app's map/yaml is never offered as current.
- `choose_app(..., current=None)` and `choose_map(..., current=None)`: mark it, and
  make Enter take it. For a map, the cooked copy wins the default over the library one
  (same map, no re-import).
- `run_profile._fmt` names the scenario file from the path — app-owned / generated /
  your variant in `~/.fixs/apps` / your variant outside it — instead of from the
  two-valued scope. `config_scope` itself is unchanged: two values is what the
  behaviour needs (what the generator may overwrite, what a map change invalidates).

## Checked against

FIXS build v0.9.0-alpha (`072ec889`), driving the real `~/.fixs/run_profiles.json`:
the mark and Enter land on the setup's own yaml; the out-of-tree, deleted-yaml and
app-changed paths behave as described; all four saved setups render the correct scope.

### LINKED ISSUE #290: [MAINTENANCE] Automate the issue lifecycle across the dev -> main branch model
**Task Summary**

Automate the issue lifecycle for the `dev_v0.X.0` -> `main` branch model: declare closing intent once at PR time, mark an issue as fixed-but-unshipped when it lands on `dev`, close it when it reaches `main`, and sweep for merged PRs whose commits never arrived anywhere.

**Why is this needed?**

GitHub's native auto-close only fires for a PR whose base **is** the default branch. Every FIXS feature PR targets `dev_v0.X.0`, so the link is never created at all -- not created-and-waiting. Measured on the merged-PR set:

```
PR#265 base=dev_v0.9.0  body='Closes #264'  closingIssuesReferences=[]
PR#268 base=dev_v0.9.0  body='Fixes #261'   closingIssuesReferences=[]
PR#274 base=dev_v0.9.0  body='Closes #272'  closingIssuesReferences=[]
PR#259 base=main        body='Closes #70'   closingIssuesReferences=[70,223,238]
```

The release PRs that do target `main` (#271, #277, #280, #282) carry no keywords, and the commits they bring use the `fix(#261):` / `(#276)` title convention -- a reference, not a keyword. So nothing fires at release time either. Result: seven issues were fixed, shipped to `main`, and left open (#127, #254, #258, #261, #264, #272, #276).

A second, quieter failure has no detector at all. PR #267 merged into `bug/254_spectator_frame_lag` at `12:51:15Z`; PR #256 had merged that branch into `dev` at `12:50:35Z`, forty seconds earlier. Commit `3f08d945` is reachable only from two dead branches -- GitHub still shows #267 as *Merged*, and the #266 fix has never executed anywhere.

**Scope of Work**

Intent is declared once, by the author, at the only moment anyone knows it. Across 73 merged dev-targeted PRs only 24 carry a closing keyword; 34 use a bare title reference that means "closes" for #269/#258 and "partial" for the twelve #191 PRs -- identical syntax, opposite meaning, unrecoverable after the fact.

- `.github/scripts/pr_intent.py` -- parse a PR body into `Closes` / `Part of` / `No issue`, cross-repo aware. Pure, no network.
- `.github/scripts/close_landed.py` -- resolve a push range to PRs via `/commits/{sha}/pulls` (strategy-agnostic: survives squash, rebase, and the cherry-picks already in `main`), then label on dev-merge and close on main-arrival.
- `.github/scripts/orphan_check.py` -- assert every merged PR's merge commit is reachable from `main` or a `dev_*` tip.
- `.github/workflows/pr-intent.yml` -- PR gate requiring one declaration.
- `.github/workflows/issue-lifecycle.yml` -- `landed-on-dev` label on dev-merge; close + unlabel on push to `main`.
- `.github/workflows/orphan-check.yml` -- weekly sweep + `workflow_dispatch`.
- `.github/pull_request_template.md` -- name the three declarations.
- `tests/Python/unit/test_pr_intent.py` -- parser tests over the real historical PR shapes.

**Risks or Dependencies**

- Cross-repo closing (`FIXS_Applications#N`, 6 PRs) cannot use `github.token` -- it is scoped to this repo, and GitHub keywords never close cross-repo regardless of branch. The job reports these for manual close; adding a PAT secret with `issues:write` on the sibling repo upgrades it to automatic with no code change.
- `close_landed` must not close on bare `#N` title references. Twelve #191 PRs would take the umbrella issue down mid-flight.
- Historical backlog: the 34 keyword-less PRs stay ambiguous. The gate fixes forward only.

yunlishao-vibe:
## Summary

GitHub honours a closing keyword only when the PR base **is** the default branch. Every FIXS feature PR targets `dev_v0.X.0`, so the link is never created — not created-and-waiting:

```
PR#265 base=dev_v0.9.0  body='Closes #264'  closingIssuesReferences=[]
PR#268 base=dev_v0.9.0  body='Fixes #261'   closingIssuesReferences=[]
PR#259 base=main        body='Closes #70'   closingIssuesReferences=[70,223,238]
```

The release PRs that do target `main` carry no keywords, and the commits they bring use the `fix(#261):` / `(#276)` title convention — a reference, not a keyword. So nothing fires at release time either. Seven issues were fixed, shipped, and left open.

This PR makes the close happen when the code actually reaches `main`.

### Intent is declared once, by the author

It cannot be recovered afterward. Of 73 merged dev-targeted PRs, 24 carry a closing keyword and 34 carry only a bare `#N/` title reference — which meant *"finishes it"* on PR #269 (issue #258) and *"one of twelve increments"* across the #191 series. Identical syntax, opposite meaning. So `pr-intent.yml` requires one of three lines in the body, and nothing downstream ever guesses:

```
Closes #N            this PR finishes it
Part of #N           incremental work; N stays open
No issue: <reason>   deliberate (release sync, merge plumbing)
```

### The three pieces

| file | trigger | what it does |
|---|---|---|
| `pr-intent.yml` | PR to `dev*`/`main` | fails unless the body declares an intent |
| `issue-lifecycle.yml` | PR merged to `dev*` | labels the issue `landed-on-dev`, leaves it **open** |
| | push to `main` | closes it, drops the label |
| `orphan-check.yml` | weekly | merged PRs whose commits reached no shipping branch |

`landed-on-dev` gives the fixed-but-unshipped state a representation it did not have. It is the machine version of the note hand-written on #264 ("leaving this open deliberately… it will close on its own"), which was also wrong about the mechanism. The label set is queryable, and it is the changelog for the next release sync.

### Two design decisions worth reviewing

**Commit → PR goes through `/commits/{sha}/pulls`, not subject parsing.** `main` already contains cherry-picks of dev commits under different SHAs (`f24d0c50` vs `5bb44a70`); the API is correct for squash, rebase and merge alike.

**Declarations are line-anchored.** Found by the dry run: PR #228's prose *"the exact signal-side analogue of the vehicle bug fixed in #176"* parsed as a closing declaration, against a repo named `in` (an optional `\s*` let the preceding English word be read as the repo part). GitHub scans anywhere in a body, but GitHub also previews the link and only acts on the default branch. This runs unattended and a false close is silent data loss, so it demands the deliberate shape the template asks for. Both failures are regression-tested.

### Nothing clones the repo

Commit ranges and ancestry use the compare API. FIXS packs ~840 MiB of history — the same reason `bundle-guard.yml` avoids `fetch-depth: 0`. Each job sparse-checks out `.github/scripts` only.

## Related Issues
Closes #290

## Environment
- Python version: 3.10 (`realsim_dev`); runners are `ubuntu-latest`

## Checklist
- [x] Tests pass locally
- [x] Issue linked above

## Additional Notes

**Validated against live data before merge**, by running the scripts locally against this repo:

- `close_landed.py --dry-run close-main --before 750252a2 --after 2c60d97f` → 41 commits, 16 PRs, 5 closes, and it correctly abstained on PR #269/#258 and PR #257/#191, reporting them instead of guessing.
- Then run for real: **#254, #261, #264, #272, #276 closed.** #127 and #258 were closed by hand — #132 predates the scoped range, and #269 declares nothing.
- `orphan_check.py --days 30` → swept 72 merged PRs, found exactly one orphan with no false positives: **PR #267**, merged into `bug/254_spectator_frame_lag` at 12:51:15Z, forty seconds after PR #256 merged that branch into dev at 12:50:35Z. Commit `3f08d945` is reachable only from two dead branches, so the #266 fix has never executed anywhere while GitHub still shows the PR as Merged. #266 stays open and now has a detector.

Three bugs the dry run caught and that are fixed here: the `in #176` parse above, `subprocess` decoding with the Windows locale codec (cp1252 dies on the em-dashes in PR titles), and success being read from stdout when `gh issue close` writes its confirmation to stderr — which marked every successful close as failed and skipped the label removal behind it.

**Do not backfill past `750252a2`.** Older PR bodies predate the convention and over-claim: #136 says `Closes #129` while `DriverModel_FIXS_Common.h` still carries eight `TODO #129` stubs, and #135 says `Closes #131` while that issue's version-tracking half is the open #279. Those declarations are inert — the workflow only ever sees new pushes — and a deep backfill is the one thing that would act on them, so the range input is documented and `workflow_dispatch` defaults to dry-run.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -13,10 +13,25 @@ What this PR does and why. Use your own sub-headings -- narrative structure is
 fine, and usually better than a checklist for anything non-trivial.
 
 ## Related Issues
-Closes #___
+<!--
+REQUIRED -- the `PR intent` check fails without one of these three lines.
 
-Sibling repos, if applicable: `FIXS_Applications#__`, `Digital-Twin-Library#__`,
-`ProprietaryFiles#__`.
+  Closes #___            this PR finishes the issue
+  Part of #___           incremental work; the issue stays open
+  No issue: <reason>     deliberate (release line sync, merge plumbing)
+
+Say it here even though the title already carries the number. A bare `#N/`
+title reference means "finishes it" on some PRs and "one of twelve increments"
+on others, and nothing downstream can tell them apart. Because this PR targets
+a dev branch, GitHub never creates the closing link itself -- the automation
+reads this line instead, labels the issue `landed-on-dev` at merge, and closes
+it when the code reaches `main`. (#290)
+
+Sibling repos take the qualified form: `Closes FIXS_Applications#21`. GitHub
+cannot close those from a keyword at all, so the automation handles them
+separately.
+-->
+Closes #___
 
 ## Environment
 Only the tools you actually tested against -- leave the rest blank:
--- a/.github/scripts/close_landed.py
+++ b/.github/scripts/close_landed.py
@@ -1,0 +1,435 @@
+#!/usr/bin/env python3
+"""Drive an issue through the dev -> main branch model.
+
+Two subcommands, one per transition:
+
+  label-dev --pr N        A PR merged into dev_v0.X.0. Its ``Closes`` issues
+                          are fixed but NOT shipped, so they stay open and get
+                          the `landed-on-dev` label. The label set is then, at
+                          any moment, exactly the fixed-but-unreleased backlog
+                          -- and the changelog for the next release sync.
+
+  close-main --before A --after B
+                          Commits arrived on main. Resolve them to the PRs that
+                          carried them, read each PR's declared intent, close
+                          the issues, drop the label.
+
+Why the resolution goes commit -> PR through the API rather than parsing
+"Merge pull request #N" out of subjects: main's history already contains
+cherry-picks of dev commits under different SHAs (f24d0c50 vs 5bb44a70), and
+nothing guarantees a release stays a true merge. ``/commits/{sha}/pulls``
+answers "which PR carried this" for squash, rebase and merge alike.
+
+Nothing here clones the repo. FIXS packs ~840 MiB of history (see the note in
+bundle-guard.yml about fetch-depth), so ancestry and commit ranges are resolved
+with the compare API instead of git. The workflow sparse-checks out only this
+directory.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import subprocess
+import sys
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from typing import Iterable, Optional
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+from pr_intent import IssueRef, parse_intent  # noqa: E402
+
+# PR titles in this repo carry em-dashes and arrows. The runner is UTF-8, but a
+# local run on a Windows console is cp1252 and would die on the summary rather
+# than on anything that matters.
+for _stream in (sys.stdout, sys.stderr):
+    if hasattr(_stream, "reconfigure"):
+        _stream.reconfigure(encoding="utf-8", errors="replace")
+
+LANDED_LABEL = "landed-on-dev"
+LANDED_LABEL_COLOR = "0e8a16"
+LANDED_LABEL_DESC = "Fixed on a dev branch; closes when it reaches main"
+
+# The compare API returns at most 250 commits per page. A release sync is tens
+# of commits, but a first push or a long-delayed sync can exceed it, and
+# silently dropping the tail would silently skip issues.
+COMPARE_PAGE = 250
+MAX_COMPARE_PAGES = 20
+
+# Concurrency for the commit -> PR resolution. Kept modest: GitHub's secondary
+# rate limiter reacts to burst concurrency, not just total request count.
+API_WORKERS = 8
+
+ZERO_SHA = "0" * 40
+
+
+class GhError(RuntimeError):
+    pass
+
+
+# Set by --dry-run. Reads still go out; writes are printed instead of sent, so
+# a backfill over historical commits can be audited before it touches anything.
+DRY_RUN = False
+
+_WRITE_VERBS = {"POST", "PATCH", "PUT", "DELETE"}
+
+
+def _is_write(args: tuple) -> bool:
+    if args[:1] == ("issue",) and args[1:2] and args[1] in (
+            "close", "comment", "edit", "create"):
+        return True
+    return any(a in _WRITE_VERBS for a in args)
+
+
+def gh_ok(*args: str) -> bool:
+    """Run a gh write and report whether it succeeded.
+
+    Success is the exit code, never stdout. `gh issue close` and `gh issue
+    edit` print their confirmation to STDERR and leave stdout empty, so testing
+    the returned text marks every successful close as a failure -- which also
+    silently skips the label removal that follows it.
+    """
+    if DRY_RUN:
+        print(f"  [dry-run] gh {' '.join(args)}")
+        return True
+    proc = subprocess.run(("gh",) + args, capture_output=True, text=True,
+                          encoding="utf-8", errors="replace")
+    if proc.returncode != 0:
+        warn(f"gh {' '.join(args[:3])} failed: "
+             f"{((proc.stderr or proc.stdout) or '').strip()}")
+        return False
+    return True
+
+
+def gh(*args: str, check: bool = True) -> str:
+    """Run the gh CLI and return stdout."""
+    if DRY_RUN and _is_write(args):
+        print(f"  [dry-run] gh {' '.join(args)}")
+        return "dry-run"
+    # encoding is pinned rather than left to the locale: GitHub bodies in this
+    # repo are full of em-dashes and box-drawing characters, and the Windows
+    # default (cp1252) raises UnicodeDecodeError on them, which surfaces as a
+    # dead reader thread and a None stdout rather than as an error.
+    proc = subprocess.run(("gh",) + args, capture_output=True, text=True,
+                          encoding="utf-8", errors="replace")
+    if proc.returncode != 0:
+        message = ((proc.stderr or proc.stdout) or "").strip()
+        if check:
+            raise GhError(f"gh {' '.join(args)} failed: {message}")
+        return ""
+    return proc.stdout or ""
+
+
+def gh_json(*args: str, check: bool = True):
+    out = gh(*args, check=check)
+    if not out or not out.strip() or out == "dry-run":
+        return None
+    return json.loads(out)
+
+
+def notice(msg: str) -> None:
+    print(f"::notice::{msg}")
+
+
+def warn(msg: str) -> None:
+    print(f"::warning::{msg}")
+
+
+def summary(lines: Iterable[str]) -> None:
+    """Append to the job summary so a run is auditable without reading logs."""
+    path = os.environ.get("GITHUB_STEP_SUMMARY")
+    text = "\n".join(lines) + "\n"
+    print(text)
+    if path:
+        with open(path, "a", encoding="utf-8") as fh:
+            fh.write(text)
+
+
+# --------------------------------------------------------------------------
+# GitHub reads
+# --------------------------------------------------------------------------
+
+def compare_commits(repo: str, before: str, after: str) -> list[str]:
+    """SHAs introduced by ``before..after``, via the compare API (no clone)."""
+    shas: list[str] = []
+    for page in range(1, MAX_COMPARE_PAGES + 1):
+        payload = gh_json(
+            "api",
+            f"/repos/{repo}/compare/{before}...{after}"
+            f"?per_page={COMPARE_PAGE}&page={page}")
+        if not payload:
+            break
+        batch = payload.get("commits") or []
+        shas.extend(c["sha"] for c in batch)
+        total = payload.get("total_commits", len(shas))
+        if len(shas) >= total or len(batch) < COMPARE_PAGE:
+            break
+    else:
+        warn(f"compare {before}...{after} exceeded "
+             f"{COMPARE_PAGE * MAX_COMPARE_PAGES} commits; the tail was not "
+             f"examined and issues it carried will stay open.")
+    return shas
+
+
+def prs_for_commit(repo: str, sha: str) -> list[int]:
+    payload = gh_json("api", f"/repos/{repo}/commits/{sha}/pulls", check=False)
+    return [p["number"] for p in payload] if payload else []
+
+
+def prs_for_commits(repo: str, shas: list[str]) -> list[int]:
+    """PR numbers carrying any of ``shas``, in first-seen order.
+
+    One API call per commit, run on a small thread pool: the calls are pure
+    I/O, and serially this is ~1 s per commit, which turns a backfill over a
+    few hundred commits into minutes. Reads only, so concurrency is safe;
+    ordering is restored from the input list rather than completion order so
+    the run is reproducible.
+    """
+    results: dict[str, list[int]] = {}
+    with ThreadPoolExecutor(max_workers=API_WORKERS) as pool:
+        futures = {pool.submit(prs_for_commit, repo, sha): sha for sha in shas}
+        for done, future in enumerate(as_completed(futures), 1):
+            sha = futures[future]
+            try:
+                results[sha] = future.result()
+            except Exception as exc:  # a single commit must not sink the run
+                warn(f"could not resolve PRs for {sha[:8]}: {exc}")
+                results[sha] = []
+            if done % 50 == 0:
+                print(f"  resolved {done}/{len(shas)} commits")
+
+    ordered: list[int] = []
+    for sha in shas:
+        for number in results.get(sha, []):
+            if number not in ordered:
+                ordered.append(number)
+    return ordered
+
+
+def pr_details(repo: str, number: int) -> dict:
+    return gh_json("api", f"/repos/{repo}/pulls/{number}") or {}
+
+
+def issue_state(ref: IssueRef, number: int) -> Optional[str]:
+    payload = gh_json("api", f"/repos/{ref.repo}/issues/{number}", check=False)
+    if not payload:
+        return None
+    # The issues endpoint also serves PRs; never act on one.
+    if "pull_request" in payload:
+        return "pull_request"
+    return payload.get("state")
+
+
+# --------------------------------------------------------------------------
+# GitHub writes
+# --------------------------------------------------------------------------
+
+def ensure_label(repo: str) -> None:
+    existing = gh_json("api", f"/repos/{repo}/labels/{LANDED_LABEL}",
+                       check=False)
+    if existing:
+        return
+    gh("api", "-X", "POST", f"/repos/{repo}/labels",
+       "-f", f"name={LANDED_LABEL}",
+       "-f", f"color={LANDED_LABEL_COLOR}",
+       "-f", f"description={LANDED_LABEL_DESC}", check=False)
+
+
+def add_label(ref: IssueRef) -> bool:
+    return gh_ok("issue", "edit", str(ref.number), "--repo", ref.repo,
+                 "--add-label", LANDED_LABEL)
+
+
+def remove_label(ref: IssueRef) -> None:
+    # Absent label -> 404, which is the expected case for anything that landed
+    # before the label existed. Not an error worth surfacing.
+    if DRY_RUN:
+        print(f"  [dry-run] remove {LANDED_LABEL} from {ref.slug}")
+        return
+    subprocess.run(
+        ("gh", "api", "-X", "DELETE",
+         f"/repos/{ref.repo}/issues/{ref.number}/labels/{LANDED_LABEL}"),
+        capture_output=True, text=True, encoding="utf-8", errors="replace")
+
+
+def comment(ref: IssueRef, body: str) -> bool:
+    return gh_ok("issue", "comment", str(ref.number), "--repo", ref.repo,
+                 "--body", body)
+
+
+def close_issue(ref: IssueRef, body: str) -> bool:
+    # Comment first: if the close then fails, the issue still carries the
+    # explanation, which is recoverable. The reverse leaves a bare close.
+    if not comment(ref, body):
+        return False
+    return gh_ok("issue", "close", str(ref.number), "--repo", ref.repo,
+                 "--reason", "completed")
+
+
+# --------------------------------------------------------------------------
+# label-dev
+# --------------------------------------------------------------------------
+
+def cmd_label_dev(args) -> int:
+    repo = args.repo
+    pr = pr_details(repo, args.pr)
+    if not pr:
+        warn(f"PR #{args.pr} not found.")
+        return 0
+
+    intent = parse_intent(pr.get("body") or "", repo)
+    if not intent.closes:
+        notice(f"PR #{args.pr} declares no closing issue "
+               f"({len(intent.parts)} 'Part of'); nothing to label.")
+        return 0
+
+    ensure_label(repo)
+    base = pr.get("base", {}).get("ref", "?")
+    sha = (pr.get("merge_commit_sha") or "")[:8]
+    rows = []
+    for ref in intent.closes:
+        if not ref.is_local_to(repo):
+            rows.append(f"- {ref.slug} - sibling repo, not labelled here")
+            continue
+        state = issue_state(ref, ref.number)
+        if state != "open":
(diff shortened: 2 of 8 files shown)


yunlishao-vibe:
Closing unmerged — scope call: the issue lifecycle stays manual for now.

Recorded on #290 so the measurements are not lost if this is revisited.

Issue #177, opened 2026-06-23
[BUG] SUMO-CM co-sim slows superlinearly — SUMO TraCI context-subscription cost (~7.5 ms/active vehicle/step) dominates, not the lib or CM traffic engine


yunlishao:
**Describe the bug**

The SUMO ↔ CarMaker co-simulation runs **far below real time, worsening as background traffic spawns in**, while the VISSIM ↔ CarMaker probe on the *same* scenario runs fast — even though **both drive the identical CarMaker-side `VirtualEnvironment.lib`**. The difference is upstream of the lib: `TrafficLayer` talks to SUMO via **libsumo/TraCI**, but to VISSIM via **DSProxy**.

Measurement localizes the dominant cost to the **SUMO TraCI context-subscription path**, not the lib and not the CarMaker traffic engine. SUMO's own per-step TraCI processing time grows roughly linearly with the active vehicle count, at **~7.5 ms per active vehicle per step** — about **1000× larger** than the CarMaker-core per-object cost (~7.4 µs/object) that #168 already measured and ~2000× the lib's flat ~3.8 µs/step.

A second symptom appeared in the same run and may share this root cause: the ego suddenly stops on the road, **rotates/spins in place, then leaves the road** (~t=133 s at a corner). See *Secondary / open question*.

**To Reproduce**

1. `cd tests/Sumo/Probes/TrafficLayer_SUMO_CMoffice`
2. Run `run_sumo_cm_demo.bat` (or headless `python verify_sumo_cm.py`) — the #174 SimpleLoop scenario, ego-free background demand, 40 looping vehicles.
3. Watch SUMO-gui's status bar (or `_logs/sumo.log`): the per-step **`TraCI: N ms`** climbs as vehicles enter, and the real-time factor collapses.

**Evidence / logs** (from `_logs/sumo.log`):

| SUMO step | active veh | TraCI ms/step |
|---:|---:|---:|
| 10  | 3  | 19  |
| 40  | 9  | 65  |
| 70  | 15 | 96  |
| 100 | 20 | 154 |
| 130 | 27 | 205 |

≈ 7.5 ms/active-vehicle/step, monotonically increasing.

**Localization (already done — keep this ticket surgical to the SUMO/TraCI layer):**

[...]

Issue #86, opened 2025-10-24
[FEATURE] Skip synchronization during warmup period to speed up XIL testing


yunlishao:
**Is your feature request related to a problem?**

XIL testing has to simulate minutes of background traffic before the ego is on the road. The question this issue opened with was whether `SimulationMode` 1 and 4 already skip synchronisation during that period, or only delay it.

**Findings (investigated, do not re-derive)**

They do skip it — the fast-forward already exists — but it is unreachable behind a broken interface and it does not overlap XIL start-up.

- **Mode 4/5** → one `libtraci::Simulation::step(T)`. SUMO runs to the target time internally with zero client I/O. This is a genuine fast-forward. `SimulationModeParameter` is **absolute** simulation time, not a duration.
- **Mode 1/2** → per-step `step()` + `Vehicle::getIDList()`, then `continue` before every client send/recv.
- **The bitfield is fiction.** The code tests `mode == 4 || mode == 5` and `mode == 1 || mode == 2` as *identical* branches, so 5 is a duplicate of 4 rather than "4 combined with 1", and 3/6/7 silently mean 0.
- **VISSIM is broken both ways.** `runSimulation` and `checkIfEgoExist` return 0 for VISSIM while their callers proceed as if they had succeeded: mode 4 is a silent no-op, mode 1 hangs forever. Separately, the DriverModel DLL has its *own* warm-up (`DriverModel_FIXS_Common.h`, `ENABLE_WARMUP = true`, hardcoded, with a comment saying it should be driven by `SimulationMode`), which neither knows about nor is known to TrafficLayer.
- **`checkIfEgoExist` is wrong.** It returns from inside the first loop iteration on both branches, so only the first subscribed id is ever checked, and it falls off the end with no return value when the subscription list is empty.
- **The warm-up waits for the XIL instead of overlapping it.** `SocketHelper::initConnection` does bind + listen + a blocking accept loop as one step, and runs before the main loop — so TrafficLayer will not begin warming up until CarMaker, CARLA and the application layer have all connected, and they then sit blocked through it.

**Measurements** (MLK scenario, 28800 → 28985, 1850 steps at the 0.1 s feed):

| | wall clock | per step |
| --- | --- | --- |
| batch `step(T)` | 2.70 s | 1.46 ms |
| tick loop + ego poll | 3.29 s | 1.78 ms |
| tick loop, no poll | 2.80 s | 1.51 ms |

[...]

Issue #41, opened 2025-08-15
[MAINTENANCE] Design milestones and associate with release version numbers


yunlishao:
**Task Summary**  
Design milestones and associate each milestone with issues, date, and release version

yunlishao:
This issue has been fully resolved:

**Implementation:**
- PR #68 implemented the automatic version management system (merged commit f92e045)
- `scripts/generate_version.ps1` auto-generates `RealSimVersion.h` from git tags during build
- Pre-Build Event configured in TrafficLayer.vcxproj for all configurations

**Milestones:**
- Active milestones: 0.7.1, 0.7.2, 0.8.0, 0.9.0
- Git tags: v0.6.0, v0.7.0

All tasks from Phase 1-3 are complete. Closing this issue.

PR #305, opened 2026-08-11
L2: the traffic simulator inserts the ego, Carla takes its dynamics over (#86)

yunlishao-vibe:
> **Ready for review.**
>
> **Where to spend attention.** Sections 0-7 are durable: the config surface, the
> traffic-simulator side, the Python backend, and the route helpers. Section 8's
> adapter (`_Map`, `_World`, `Waypoint`, `_LaneMarking`, and the geometric
> obstacle sweep) is an **interim layer scheduled for replacement** -- it
> reimplements a road network out of the route polyline, when the bridge process
> is holding a live CARLA connection and could forward the question to the real
> map instead. Stage 2/3 replaces it with a FIXS world-query API, and that change
> is mostly a subtraction. Review section 8 for behaviour, not for its internals.
>
> This PR is infrastructure: it makes it possible for a Python controller to
> occupy the ego's driver slot and close its loop on the CARLA plant. That is
> verified end to end (below). It does **not** claim a result about which driver
> is better -- that comparison is follow-up work and is not a gate on this code.
>
> Scope grew from the original ego-spawn change: the Python backend needed its
> half of the L2 path, because that is where a Python controller can run. The C++
> bridge had the deferred spawn and no controller hook; the Python backend had
> the hook (#337) and demanded `EgoSpawnPose`. Neither had both.
The ego half of the L2 work. The warm-up change that was previously on this branch has been split out into ORNL-Real-Sim/FIXS#307, which is independent and can land first.

## 0. `EgoSetup` — the externally-driven ego, described once

A vehicle whose motion something outside the traffic simulator computes is the same situation whether that something is CarMaker, an XIL plant or Carla: SUMO holds a vehicle it did not create, and whose position is written in from outside every tick. It was nevertheless described twice, and nothing stopped the two descriptions naming different vehicles:

| described twice | now |
|---|---|
| `CarMakerSetup.EgoId` / `EgoType`, `CarlaSetup.EgoId` / `EgoSumoType` | `EgoSetup.Id` / `SumoType`, which fill all four |
| `addEgoVehicle()` and `addEgoVehicleFromXY()` | both, selected by owner, in one mirror body |
| two branches in `sendToSUMO`, one per id set | `externalEgoOwnerOf()` + one body |
| `keepRoute` hardcoded `6` for CarMaker, configurable for Carla | `EgoSetup.KeepRoute` for both |
| `EgoDynamics: carla` — a value named after a backend | `virenv`, which CarMaker occupies identically |

`EgoSetup` is read **last** and writes back into every key it replaces, so the rest of the engine reads exactly what it read before and older scenarios parse unchanged. A half-migrated config that says two different things is reported rather than silently resolved.

The vocabulary drops the backend names, because the distinctions were never about backends:

```
Dynamics         traffic | virenv | xil      ("carla" still accepted)
ActuationSource  simulator | fixs | user     (old four still accepted)
```

`simulator` is the driver the environment brought with it — Carla's TM, CarMaker's IPGDriver. `fixs` is the one FIXS ships. `user` is yours, and **where it runs is not a fourth value**: name a `Controller` file and it is called in-process once per environment step; leave it unset and the pedals come off the FIXS record at the 0.1 s feed. That difference is load-bearing — a feed-side controller reads the advisory back as its own measured speed and never corrects — but it is a consequence of the `Controller` key, not a separate kind of driver.

Two places in the merged mirror still branch on the owner, and both are marked in the code. The Carla ego is injected from `(x, y)` because it has no first-step hook and its spawn position is only known from the feed; a CarMaker ego added but not yet departed is absent from `getIDList`, so injecting there too would try to create it twice. And `setSpeed` is not yet applied to the CarMaker ego, only because nobody has run CarMaker since this merged — the mechanism applies there equally.

**What CarMaker gains from sharing the path:** a configurable `keepRoute` instead of the hardcoded `6`, and the per-tick `getNextTLS` reseed. The next-signal cache (#177) locates a signal by subtracting the odometer from stored cumulative distances, which assumes the odometer only advances with distance travelled — false for any `moveToXY`-placed vehicle, and CarMaker's ego is one.

**Verified against both parsers**, since two of them is how an L2 config once ran as L0 on one side and L2 on the other:

- Python — `test_ego_setup_config.py`, 19 tests. The strongest one parses each shipped config **as it stands now** against the same file **as it stood before `EgoSetup` existed**, read straight out of git, and requires the derived values (`EgoMode`, `EgoL0Driver`, `EnableExternalControl`, `EgoId`, `EgoSumoType`, `EgoController`) to be identical. Those are what the bridge acts on, and a migration that quietly moved one would not announce itself in a run — the ego would just be driven by something else.
- C++ — `TrafficLayer` reads the migrated config (`Read configuration file success`), refuses an unknown `Dynamics` with the new message, and warns then resolves to `EgoSetup.Id` when a stale `CarlaSetup.EgoId` is left behind.
- Both solutions rebuild clean (`TrafficLayer`, `VirCarlaEnv`, `VirtualEnvironment`; exit 0).

## 1. The traffic simulator can insert the ego

`VirCarlaEnv` spawns the ego before the co-sim loop and requires `EgoSpawnPose`, so Carla is always the creator and TrafficLayer injects the result into SUMO on a single-edge dummy route. This adds the other direction — the one the CarMaker/XIL path has always used — selected by whether `EgoSpawnPose` is set, so nothing existing changes:

| `EgoSpawnPose` | who creates the ego |
|---|---|
| given | unchanged: the bridge, before the loop; SUMO gets it injected |
| absent | the traffic simulator; the bridge spawns a physics actor at the pose it reports, the first tick the ego id arrives |

With Carla as creator, the ego's entry time has to be duplicated outside the traffic scenario, and SUMO's copy of the ego gets a single-edge route — so no next traffic light, which is the only stop-bar input a signal-aware controller has.

- `mainVirCarla.cpp` — spawn + driver wiring factored into one place, callable before the loop or inside it. The deferred path deliberately does **not** tick: an out-of-band `world.Tick()` there advances Carla past the feed and desyncs every mirrored vehicle.
- `TrafficHelper.cpp` — the inject path **adopts** an ego SUMO already has. `vehicleExist` was computed at that site and never read; without it `Vehicle::add` throws "already exists" every step, and since the throw precedes the `moveToXY`/`setSpeed` in the same `try`, the Carla ego was never mirrored at all.

## 2. Do not feed the traffic simulator an ego it was told not to take

The post-tick readback published the Carla ego's record whenever `EgoMode >= 1`, without checking `EnableExternalControl`. With that flag false there is no `carlaOwnsId` path to consume it, so it fell through to the ordinary application-layer branch and was applied as a plain `setSpeed` on the traffic simulator's ego — from the highest-numbered client, so it also beat the controller that was supposed to be driving.

"Own an ego in Carla but do not feed it back" silently became "drive the traffic simulator's ego from Carla". It is what makes an `EnableExternalControl: false` control run fail to reproduce its own baseline, and it is invisible unless you diff a trajectory.

## 3. `EgoSetup.KeepRoute` and a route-replacement guard

`EgoSetup.KeepRoute` (default `6`, the previously hardcoded value; `SumoSetup.EgoKeepRoute` still parses) exposes the `keepRoute` bitmask for the mirror's `moveToXY` — really a choice of failure mode for a vehicle SUMO no longer drives: bit 1 places it at the exact position and lets it leave the network (off-road driving possible, degradation silent); bit 0 pins it to its own route and makes SUMO raise instead.

`ego_sumo_route_replaced` fires when the route was replaced *by the mirror*, which is not the same event as a route change. Changing a route is a normal application action — route guidance is a CAV function — so raising on "the routeId differs" would teach users to ignore it. SUMO's replacement "consists of that edge only", which no deliberate reroute produces. Ordinary looping cannot reach it either: `<route repeat="n">` is expanded at load (measured: one routeId holding 1470 edges on the MLK arterial).

## 4. The ego driver needs an integral term

The speed law is a *tracking* law, so a controller pulling away from a stop bar asking ~0.15 m/s gets `0.25*0.15 + 0.15 = 0.19` throttle. Probed at the ego's own spawn with the same blueprint (1845 kg): **0.45 throttle → 5.7 m in 4 s (0.34 m/s); 0.75 → 44.2 m (2.98 m/s)**. The response has a knee, so a fixed floor is the wrong instrument — the right value differs per vehicle, grade and surface. Integrating the deficit finds it. Effect on the MLK L2 run: 0.54 m travelled → 5513 m.

## 5. Observability

The bridge reports what the in-Carla driver applied on the ego's own actuation fields, `DataLogger` learned those three names, and the handover prints a 5 s trace. "It was told to move" and "it moved" were previously indistinguishable in every artifact a run produced, which is most of why the first stall took a day to find.

## 6. `sumo_route_points.py` — the ego route as a polyline, from the map's own route

`sumo_ego.py` puts the ego on a route in SUMO, and a route in SUMO is a list of edge ids. Nothing in Carla can steer to an edge id, so an L2 scenario needs the same route as `(x, y)` points — for the pursuit driver here, and for a Python controller built on Carla's `BasicAgent`. That second copy had no owner: it was generated by the MLK application out of a hardcoded edge list, which is the pattern `sumo_ego`'s own docstring exists to condemn, and the one the U-turn work had already removed from the SUMO side.

This derives it from the same place `sumo_ego` does — the route id, in the bundle's own demand, on the network that bundle names. The lane-connection walk, the junction interiors and the chain search the application had worked out come across unchanged; only the **input** changes.

`--check` compares a yaml's stamped fingerprint of `(network, route id, edges)` against the bundle. The two artifacts fail apart silently: regenerate one and forget the other and SUMO drives one corridor while Carla steers another, and the first symptom is not a visible swerve but the ego's next-signal fields going empty — which an eco-driving controller reads as "no signal in sight" and plans against.

Verified on `mlk_notexture_uturn` `route1`: 76 edges, 361 points, 5388 m, zero lane changes, and the polyline reproduces both `sumo_uturn` turnarounds (42+96+42 deg over 338 m, minimum radius ~87 m). `--check` accepts that map and rejects `mlk_no_signal`'s `route1` (`a99337d0` vs `172831c9`).

## 7. The Python backend can run this scenario, and a controller in the driver slot

`_setUpEgo` becomes `_bringUpEgo(..., sp, settleOutOfBand)`, the peer of the C++
`bringUpEgo` lambda, callable from the same two places: before the loop at a
configured `EgoSpawnPose`, or from inside it at the pose the traffic simulator
just reported. `settleOutOfBand` is why it is one function and not two — before
the loop those extra ticks settle the ego and pay the traffic manager's one-time
map build; inside the loop they must not happen, because an out-of-band tick
advances Carla past the feed and desyncs every mirrored vehicle.

`EgoActuationSource` gains `embedded`. Setting `EgoL0Driver` directly would have
worked and left one config speaking a different language from its neighbours. The
header now also says why the two controller placements are not interchangeable: an
external client is served at the 0.1 s feed and sees the ego record as the traffic
simulator left it — including `speed`, which under L2 carries the eco advisory,
not the measured speed. A controller closing a speed loop there reads back its own
setpoint, and its error is ~0 by construction.

Three further gaps, each found by running it and each silent:

- **`ConfigHelper.py` had no `EgoDynamics` / `EgoActuationSource` derivation.** Those
  are what a scenario sets; `EgoMode` and `EgoL0Driver` are what the bridge reads,
  and only the C++ helper derived one from the other. So the Python bridge read the
  declared vocabulary as `EgoMode 0` and quietly ran an L0 scenario while
  TrafficLayer, reading the same file through the C++ helper, ran an L2 one. The ego
  was never taken over and the controller never called. It also never parsed
  `EgoController` at all.
- **`EgoL0Driver: Embedded` fell through to the traffic manager**, which would have
  steered to its own plan with the controller's pedals applied on top.
- **The embedded call site asked for the ego before the first receive.**
  `fixs.vehicle.get` raises rather than answering empty — deliberately, since "no
  tick yet" and "a tick with no ego" differ — and the core skips the recv at
  simTime 0, so the first pass threw and the bridge exited.

`DataLogger.logVehicle` also takes `idOverride`: the `ego_sumo` row was produced by
`copy.copy` of the received record and assigning `.id`, which a fixs record refuses.
Unreachable until an L2 ego existed on this backend.

## 8. BehaviorAgent, and the adapter surface it needs

`agents/` was vendored as 8 of upstream's 11 files, and `VERSION.txt` said "files
8, all BYTE-IDENTICAL" without saying it was a subset. The three absent were
`behavior_agent.py`, `behavior_types.py` and `constant_velocity_agent.py`, and
that omission was load-bearing: `BasicAgent`, the only agent vendored, has **no
longitudinal following model**. Its `run_step` asks "is anything inside
`5 + speed` metres" and if so calls `add_emergency_stop` — zero throttle, full
brake, no gap or relative-speed term.

Measured with it in the driver slot on the MLK U-turn corridor: 27 impacts in
3189 s, and in **11 of them the hazard had been flagged continuously for the
preceding 2 s**, closing to 3.0 m. The agent saw the obstacle and had no response
available but to slam on. The sensing was not the problem — the sweep is handed a
median of 196 vehicles per step and raises a hazard on 10.5% of them.

So `behavior_agent.py` and `behavior_types.py` are now vendored (md5-identical to
upstream) and `VERSION.txt` states the subset and why it matters.

Two integration points were needed:

- `_vehicle_obstacle_detected` returns an obstacle object rather than `None`.
  `BasicAgent` discarded that element; `BehaviorAgent` measures it — `run_step`
  subtracts its half-extents to turn a centre distance into a gap, and
  `car_following_manager` reads its speed to band on time-to-collision.
- The adapter grew a bounding box, a speed limit, and waypoint
  `lane_width` / `is_junction` / `lane_id` / lane markings. The markings report
  `LaneChange.NONE` deliberately: that switches tailgating off, and it must be
  off, because this ego drives the traffic simulator's route and `_tailgating`
  would call `set_destination` to re-plan through a `GlobalRoutePlanner` whose
  topology is empty here.

**The road's speed limit and the eco advisory are kept separate, and that is not
cosmetic.** `BehaviorAgent` scales its own look-ahead off the speed limit —
`collision_and_car_avoid_manager` searches `max(min_proximity_threshold,
speed_limit / 3)` metres. Feeding an eco advisory of 5–8 m/s there collapses it
to the 10 m floor, about 1.25 s of headway at 8 m/s, so the first thing it ever
notices is already inside `braking_distance`. Measured while it was wired that
way: a leader 13 m ahead at 3 m/s went undetected entirely, and the agent went
from clear to full brake between 11 m and 9 m — reproducing the `BasicAgent`
failure with new code. `get_speed_limit()` therefore returns the road limit; the
advisory arrives via `behavior.max_speed`, which every branch of `run_step` reads.

## Testing

- Builds clean, Release x64.
- Ran end to end on the MLK arterial with headless CARLA: SUMO inserts the ego at its own depart time, the bridge adopts it, and the pursuit driver takes it along the corridor tracking the eco controller's advisory.
- With `EnableExternalControl: true`, the Carla ego and SUMO's ego agree to a **median 0.02 m, p95 0.03 m** over 6501 feeds, and the ego keeps its next traffic signal on **100%** of controlled ticks.
- Ruled out with data during bring-up: no mirrored twin of the ego (0 of 1.23 M pose-log rows), spawn yaw correct (23.4° vs the road's 23.4°), elevation correct, and the road surface under the ego flat to ±0.00 m where it once launched.

### The U-turn map (2026-09-07)

Re-run on `mlk_notexture_uturn`, whose corridor ends in real `sumo_uturn` turnarounds instead of SUMO's ~5 m point reversal. **The pursuit driver drives the turnaround.** Heading sweeps 75.7 deg -> 14.9 -> -48.2 -> -88.7 -> 254.8 — a full 180 deg reversal — at 9.3-9.7 m/s with z flat at 215.1 m throughout. On the old network the ego left the road at that same point (x~2300, t~398 s), because a 1.6 m point-turn radius is not trackable; these loops are ~87 m minimum radius, 1.3 m/s2 lateral at the speed the ego arrives at.

Over the full run the pursuit driver went **18.09 km — three complete laps and six turnarounds** (east at t = 402 / 1267 / 2368 s, west at 798 / 1847 / 2814 s) without leaving the road: route deviation **median 0.02 m, p95 1.49 m, worst 7.5 m** across 3171 s, zero ticks at an implausible z. The run ended at 3171 s of 4200 because CARLA died unprompted, its log ending mid-operation with no crash trace; not investigated, since three laps was already past what was needed.

**Open, and why this stays draft — see "Known / not settled" below: the SUMO mirror no longer holds on this map.**

### The embedded controller, on the Python backend (2026-09-08)

End to end on `mlk_notexture_uturn`: the traffic simulator inserts the ego, the
backend takes it over (`Ego 'ego' entered at t=0.10 ... spawning the physics ego`),
and the controller drives it.

The property that matters, joining the controller's own per-step log to the bridge
datalog by position over 982 matched steps:

| what the controller read as its speed | steps |
|---|---:|
| the CARLA plant's measured speed | **980** |
| the eco advisory | **0** |

On the external-client path this is the exact inverse — the advisory on every tick
and the plant on none, which is what made that arm's loop open.

`tests/VirEnv`: 106 passed, 1 skipped. Non-regression only — none of them reach the
ego bring-up this changes.

## Known / not settled

- **The SUMO mirror diverges on `mlk_notexture_uturn` (blocks un-drafting).** The claim above of a median 0.02 m ego/mirror agreement was measured on `mlk_no_signal`; re-measured on the U-turn map it is **median 71.7 m, p95 189.7 m**, ramping monotonically 1 m -> 286 m rather than jumping. The old baseline on the old map held 0.69 m median and stayed under 5 m until it left the road.

  It detaches at a single event, t=83.5 s, x~406: in one 0.1 s tick the ego's heading changes 102.0 deg -> -81.3 deg (183 deg) while its x moves *backwards* 405.6 -> 401.7 m and the reported speed does not change. Over the next 1.5 s z rises 205.1 -> 211.1 m and returns to 205.0, and the steer command saturates at the -0.7 clamp. The ego recovers and drives on; the mirror never re-anchors, and from there SUMO's copy runs ahead.

  What differs at that spot: this run enters it at **12.4 m/s under full braking** (advisory 6.0), where the old baseline entered the same x-range at **4.5 m/s** with z flat and steering under 0.03 — so it is at least speed-dependent, and the map's terrain there may differ too. I have not established the cause and am not guessing at one. What would settle it: the same map driven through that junction at the old baseline's entry speed, and VirCarlaEnv's own console for that window (it goes to a separate window, not the run log, so this run did not capture whether `moveToXY` was refused).
- The driver's integral has no anti-windup; it fixes launch but overshoots at cruise (peak 37 m/s on a 13.4 m/s arterial).
- **The ego's state is not coherent between Carla and SUMO, in both directions — tracked as #353, and the blocker for un-drafting.** The claim above of a median 0.02 m ego/mirror agreement was measured on `mlk_no_signal`; re-measured on the U-turn map it is **median 71.7 m, p95 189.7 m**, ramping monotonically 1 m -> 1411 m rather than jumping. It detaches at t=83.5 s, x~406, where in one 0.1 s tick the heading changes 102.0 deg -> -81.3 deg while x moves *backwards* 405.6 -> 401.7 m; z then rises 205.1 -> 211.1 m and returns, and steer saturates at the clamp. Cause not established; #353 lists the three experiments that would settle it.

  The other direction is worse for anyone building on this: with `EgoActuationSource: external`, the controller client subscribes to `type: "ego"` and receives the **traffic simulator's** record. The Carla plant's state exists only inside the bridge, so an external controller has no path to the state of the thing it drives. Measured on the Python `BasicAgent` arm — its speed equals `ego_sumo` to the last digit at every tick, never the Carla ego, so at t=9 s it believed it was at x=179.5 doing 6.65 m/s against a 6.80 m/s advisory while the vehicle was at x=226.2 doing 14.56 m/s, and its own log read as excellent tracking throughout. That arm covered 1.87 km and lost its actor at t=424 s; the in-bridge driver, which reads the Carla actor directly, covered 18.09 km on the same scenario.

  There is an ordering constraint inside this, not just plumbing: clients are served in ascending port order and each returned record replaces that id's record for later clients, the bridge sits on 440, and a controller must write pedals *before* the bridge reads them. So same-tick plant state is impossible by construction; one-tick-old is the natural answer, and today the record carries none at any age.

  An earlier revision of this section blamed mirrored traffic destroying the ego, on the reasoning that a driver which brakes ends up sitting where a teleported actor passes through. That was wrong and is withdrawn — the overspeed above is what put the ego at the back of a queue at ~14.7 m/s. The uncommanded 7.1 m displacement it then suffered is real and still unexplained, but downstream.

- The driver's integral has no anti-windup; it fixes launch but overshoots at cruise (peak 37 m/s on a 13.4 m/s arterial).
- **Mirrored traffic destroys the ego, and NOT only with `EnableExternalControl: false` (widened 2026-09-07).** The original note here said the Carla ego is invisible to SUMO when external control is off, so mirrored traffic — teleported, hence immovable to PhysX — drives through it and ejects it; in one run it fell to z = −10 km. A run with external control **on** has now done the same thing, so that scope was too narrow and a collision-level fix is needed rather than a configuration rule.

  Measured on the `EgoActuationSource: external` arm: the ego sits at exactly (260.6, 875.5) for 10.5 s with throttle 0 and a light brake — its own obstacle detection holding it for a vehicle ahead — then in one 0.1 s tick moves **7.1 m with a 76 deg heading change, still at zero throttle**. No commanded control produces that. It tumbles at t=202 s (2–3 m per tick, heading spinning, speed pinned at 18.00 m/s) and by t=424 s the actor is destroyed, with 1149 later rows reporting it at the origin.

  This is worse for a driver that STOPS. The in-bridge pursuit driver kept moving on the same scenario and was never sitting where a mirrored vehicle would pass through it — it ran 18.09 km, three laps, six turnarounds, without leaving the road. So the arm that brakes correctly for obstacles is the arm that gets destroyed, and any safety comparison between drivers is unmeasurable until this is fixed.
- The eco advisory can stall the ego at a crawl when the controller's own signal timing is wrong (e.g. behind a warm-up without #307), because the plan ramps from the measured speed.

Consumer: ORNL-Real-Sim/FIXS_Applications#41.


yunlishao-vibe:
End-to-end check of the merged mirror, on the migrated config (`318d95ab` + Applications `66e962aa`).

600 s of MLK U-turn corridor, `mlk_notexture_uturn`, py engine, embedded BehaviorAgent, CARLA tick 0.05:

- the ego is **adopted**, not recreated: `veh_type: EGO_TYPE_EXTERNAL` on every record, i.e. SUMO inserted it on its own route and the mirror took its dynamics over
- **next signal on 3000 of 3000 ego records, `control: True` on 3000** — the merged mirror preserves the route and the per-tick `getNextTLS` reseed holds. The failure this replaces is 11.2% availability, which is what an eco controller reads as "no signal in sight"
- 1835.5 m driven continuously, moving on 5129 of 6000 bridge records, mean 3.58 m/s while moving. The only two position discontinuities are at t=0.1 and t=0.2 with speed 0.00 — the spawn settle, not a mid-run teleport
- `mainVirCarla exited (0)`, graceful shutdown, no processes or held ports left behind

The config that drove this is the migrated one: `EgoSetup.ActuationSource: user` plus a `Controller` file is what selected the embedded slot and loaded the agent (`[agent] BehaviorAgent embedded: 361 route points -> 2558 after densify`).

yunlishao-vibe:
Follow-up on the review point: the duplicated fields are **deleted**, not just kept in agreement (`327c8c9f`).

The first pass left `CarMakerSetup.EgoId`/`EgoType` and `CarlaSetup.EgoId`/`EgoSumoType`/`EgoDynamics`/`EgoActuationSource`/`EgoController` in place with `EgoSetup` writing into them. That is still two descriptions with a promise they agree, and a later change can set them apart. The fields are gone; only the yaml **keys** are still read, as fallbacks in the `EgoSetup` section, so older scenarios parse unchanged. Every consumer reads `EgoSetup` now — `TrafficHelper` (5 sites), `mainTrafficLayer`, `VirEnvHelper`, the C++ bridge (2), the Python bridge (5).

A sweep for the rest found three more that were dead or duplicated: `CarlaSetup.EgoDynamicsOwner` and `EgoControl` (the #174 predecessors of `Dynamics`/`ActuationSource`, zero readers) and `CarlaSetup.EnableEgoSimulink` (a duplicate — every real reader uses the CarMaker one). `SumoSetup.EgoKeepRoute` is gone as a field too; `EgoSetup.KeepRoute` reads that key as a fallback. What stays under `CarlaSetup` is `EgoBlueprint`/`EgoSpawnPose`/`EgoRoutePoints`/`EgoRouteRepeat`/`EgoTargetSpeed` (driver and rendering behaviour, single) and `EgoMode`/`EgoL0Driver`/`EnableExternalControl` (the derived internal representation, also single).

One deliberate behaviour change, in the case where the old behaviour was already wrong: with no ego id named anywhere, the id is now inferred from the lone vehicle subscription before any default. The old per-backend defaults ("ego" for Carla) named a vehicle that did not exist whenever the subscription was called something else.

**A correction that belongs in this thread.** Commit `1fe70c43` vendored `behavior_agent.py` and its message describes the adapter changes it needs, but it committed only the two vendored files — the adapter half lived in a working bundle and never reached this repo, and syncing the repo copy over that bundle destroyed it. It is rebuilt in `327c8c9f` from that commit message and the tests: `import_carla_agents` returns `BehaviorAgent`, `_vehicle_obstacle_detected` returns an `_OtherVehicle` the following model can measure instead of `None`, `EgoAdapter` carries a bounding box and `get_speed_limit` (the ROAD's limit, never the eco advisory, which collapses the obstacle search to its 10 m floor), and `Waypoint` carries the lane-graph attributes `BehaviorAgent` reads. Those are inert — the adapter has a route polyline and no road network — which is exactly the limitation stage 2/3 removes.

Also fixed: `fixs_adapter.read_ego_id` greps `EgoId:` out of the yaml, and the migrated configs no longer have that key. It fell back to its default, which happened to match; with any other id the standalone client would have attached to the wrong vehicle.

Comments trimmed to the high-level picture throughout, with the reasoning here rather than in the source.

Verified after the change:
- `TrafficLayer` / `VirCarlaEnv` / `VirtualEnvironment` rebuild, exit 0
- 30 tests pass (19 config, 11 controller)
- 300 s on the U-turn corridor: ego adopted as `EGO_TYPE_EXTERNAL`, **next signal on 1500 of 1500 ego records**, `control: True` on all of them, 1062.6 m driven, mean 3.57 m/s while moving (3.58 before the reconstruction), `mainVirCarla exited (0)`

yunlishao-vibe:
Naming follow-up (`3edb20a9`): two keys still carried a backend into a section meant to be free of one, and they resolve in opposite directions.

`EgoSetup.SumoType` -> `EgoSetup.Type`. "The vehicle type the traffic simulator knows this vehicle by" is a generic concept — VISSIM has one too — and the name should not decide which simulator is allowed an ego. Renamed rather than moved, because it *is* an ego property. The virtual environment's model stays `CarlaSetup.EgoBlueprint`, which is genuinely Carla-specific.

`EgoSetup.KeepRoute` -> back to `SumoSetup.EgoKeepRoute`. The opposite call by the same rule: this one is not generic at all. It is the raw bitmask argument to SUMO's `moveToXY`, bit for bit, and a section describing the ego has no business holding a TraCI parameter. Both owners still read the one field, so CarMaker's hardcoded `6` is still gone — which was the point of touching it.

**Also swept the rest of the ego surface for the same problem.** What stays, and why:

- `CarlaSetup.EgoSpawnPose` — the strongest remaining candidate for `EgoSetup`: its presence or absence decides *who creates the ego*, which is ownership, not rendering, and the pose is already in the FIXS frame. Six read sites plus the deferred-spawn logic, so it is a change rather than a rename; not worth growing this diff mid-review.
- `CarlaSetup.EgoRoutePoints` / `EgoRouteRepeat` / `EgoTargetSpeed` — driver behaviour, backend-agnostic in concept. The existing comment already says they should leave the sim config for the driver application entirely, which is a better destination than `EgoSetup`; stage 3 touches the driver, so that is the moment.
- `CarlaSetup.EgoMode` / `EgoL0Driver` / `EnableExternalControl` — the derived internal representation, never written by a scenario. Single, so no duplication to fix.
- Backend names left inside `EgoSetup` are all in comments naming the concrete thing that occupies a generic slot ("Carla PhysX today", "Carla TM, IPGDriver"). Those earn their place.

Verified after the move — the risk being that `EgoKeepRoute` silently stopped reaching the mirror and fell back to `6`, which shows up as next-signal availability collapsing to ~11%:

- three solutions rebuild, exit 0; 30 tests pass; `TrafficLayer` reads the migrated config
- 300 s on the U-turn corridor: **next signal on 1500 of 1500 ego records**, `control: True` on all of them, ego adopted as `EGO_TYPE_EXTERNAL`, 1062.6 m driven, mean 3.57 m/s while moving — identical to the run before the move
- `mainVirCarla exited (0)`, no processes or held ports left behind

yunlishao-vibe:
The reasoning that used to live in the code, moved here (`9348231f`).

You were right that the previous pass was not thorough — I scanned ten files at an arbitrary run-length threshold instead of reading what the branch actually added. Doing it properly (diff against the merge base, every added comment line): **425 added comment lines, now 356**, and the ones that remain answer "why is this line here" rather than narrating the incident that produced it.

Two real bugs fell out of *reading* it rather than scanning it: a "Carla drives the ego / order matters" header had been duplicated onto the bring-up lambda where it described the wrong code, and the C++ rename was half-applied — `EgoL0Driver` was gone from the config but `useFallbackDriver` / `driveEgoFallback` survived in `CarlaBackend` and the bridge. Both are now `stepEgoDriver` / `useFixsDriver`, matching the Python side.

Here is what moved out, so it is not lost:

**`settleOutOfBand` (the ego bring-up).** Before the loop the bridge owns the clock and can tick the world to drop the car onto its tires. Inside the loop it does not: an extra `world.Tick()` there advances CARLA past the feed and desyncs every mirrored vehicle. A deferred ego is spawned as the traffic simulator inserts it — at ~0 m/s — so it settles over the loop's own next ticks instead (0.5 s at `CarlaTimeStep` 0.05).

**Why the driver line names all three.** It predates wire actuation and read "EgoDriver" or "TM", so a run with the pedals arriving over the wire — where no in-CARLA driver runs at all — announced itself as being driven by the Traffic Manager. That is the one option this scenario explicitly rejects, and the operator has no other line telling them which driver they got.

**Spectator into the same batch as the vehicles.** A standalone `SetTransform` RPC is its own message and can be applied in a different tick from the poses it is supposed to be centred on — which put the followed vehicle one step off centre on **4.7% of frames** even after the vehicles themselves were made reliable.

**The advisory on the readback path.** `lastAdvisory` is otherwise only refreshed in the owned-ego branch, which the wire-actuation mode skips. That made `speedDesired` a constant in every external run, so the standard datalog could not be used to compare an external driver against an internal one at all.

**Observing is not coupling.** The ego record is built and logged on every feed; only the push back into FIXS is gated on ownership. This matters most in the control run, where that loop is the only thing that sees the ego: logging it is how the mirror gets verified. The CARLA and SUMO rows should agree to within the ~2-tick staleness, and a real gap between them is the bug worth catching — a comparison expected to read zero is exactly the one worth recording, because nothing else would notice it drift.

**Why "had a next signal, now has none" is deliberately not guarded.** `signalLightId` only goes empty when no cached entry is still ahead of the odometer, which means either the cached list is empty — the route was replaced, which changed `routeId` and the guard above already fired — or the vehicle has passed the last signal on its route, which is ordinary. Guarding it would warn at the end of every run and teach everyone to ignore the one warning that matters.

**The next-signal cache, in full.** `getNextTLS` is O(remaining route length) and was called per vehicle per tick, so #177 cached the signals-ahead list once and located the nearest by arithmetic (`cumDist - odo`). That assumes the odometer advances monotonically with distance travelled — true for a driven vehicle, false for one placed by `moveToXY`. Measured on the closed loop: the ego's odometer freezes for several ticks at a junction and then **regresses (−17.2 m at t=29175.6)** while the CARLA ego is on the road 0.02 m off the lane centreline at 4.5 m/s. `cumDist` is seeded once and never revisited, so one regression biases every later distance permanently — median +21.7 m, worst −176 m — and the list then names a signal the ego has already passed. That is the whole divergence: it starts on the exact tick the odometer freezes (1706 of 1710 mismatches follow it) and it is not lane changes — L0's ego changes lane five times, the head follows correctly, and it never diverges in 5089 ticks.

**And the head-index match.** The fallback path runs on 84.8% of ticks (not "rare"). `getNextTLS`'s first entry was a *different* signal than the cached one on 2249 of 5511 of them, and the head was reported as e.g. 202605864 head 9 when the 9 belonged to 202587081, 81 m away. Head indices the application then rejected: 1669 of those 2249 (74.2%), against 0 of 3262 where the two agreed. A SUMO-driven ego never diverges, so L0 never saw it — 0 rejections in 6502 ticks.

**The removed probe.** `RS_EGO_ROUTE_LOG=<path>` recorded one row per exchange for `RS_EGO_ROUTE_ID` (routeId, edge count, route index, odometer, cache size, reseeded flag, resulting signal). It existed because "the ego lost its next signal" has two candidate causes that look identical downstream — the route was replaced, versus the cached list is fine but the odometer arithmetic no longer resolves — and no existing output distinguished them. A replaced route shows `nRouteEdges` collapse with `reseeded=1` on the tick the signal goes empty; an odometer problem shows the cache still populated with the signal empty and `reseeded=0`. It found the cause, so it is gone.

Verified: three solutions rebuild (exit 0), 36 tests pass.

yunlishao (approved):

Issue #322, opened 2026-08-26
Per-axle dyno torque injection into CARLA: two pedals are not enough for AWD


yunlishao:
## Context

Follow-on from #319. The XIL bench needs to inject the dyno's measured **per-axle** torque
into CARLA. For a two-motor AWD EV (the EV6 we have drive-cycle data for) that means front
and rear axle torque under independent control.

CARLA exposes no per-wheel or per-axle drive-torque command. The only per-wheel setter in
the whole server plugin is steer. `apply_control` gives one lumped throttle that PhysX
splits through a differential we cannot write to.

## What was established

**The differential is real and its constants are known.** `UWheeledVehicleMovementComponent4W`
copies PhysX's defaults verbatim and CARLA never overrides them:
`mFrontRearSplit = 0.45`, left/right 0.5 on each axle, limited-slip with bias 1.3. Measured
0.448 on the Tesla Model 3, matching. So per-axle injection is the right granularity — the
axle differentials handle left/right, and the mismatch lands on the differential model.

**Throttle and brake superpose**, and `max_brake_torque` is per wheel, so throttle plus an
asymmetric brake bias give two independent axle degrees of freedom. Measured: front share of
brake torque sweeps 0.236 -> 0.708 as the bias goes rear-only -> front-only, with total
torque and deceleration unchanged.

**Two knobs are not enough for any AWD vehicle.** The reachable set is a cone whose edge is
pinned at the differential ratio, which forces traction and regen onto *opposite* sides of
it. Every real AWD car has the same primary axle leading in both regimes, so the two spans
overlap and cannot be separated. EV6 data: traction front share spans [0, 0.49], regen
spans [0, 0.46]. Sweeping both the brake bias and a hypothetical custom differential ratio,
the best achievable split error is p99 = 336 Nm against +-1100 Nm axle torques.

**The handbrake is the missing third channel.** PhysX sums brake and handbrake per wheel
(`PxVehicleUpdate.cpp:1901`), `max_handbrake_torque` is a separate per-wheel field, and
CARLA's Model 3 already ships it rear-only `[0, 0, 1400, 1400]`. Since traction and regen are
temporally disjoint, a boolean is enough. With a rear handbrake >= 1600 Nm the split error
over the whole EV6 cycle drops to **exactly zero**.

## Open items

1. **Measure the handbrake transition penalty.** `HandbrakeInputRate` is 12/s, an ~83 ms
[...]

Issue #319, opened 2026-08-25
Investigate CARLA 0.9.16 migration: get_telemetry_data exposes per-wheel omega and torque


yunlishao:
## Why

CARLA 0.9.16 adds `get_telemetry_data` to the vehicle actor. It returns per-wheel
`omega`, `torque`, `long_slip`, `lat_slip`, `tire_load`, `long_force` and `lat_force`,
plus vehicle-level `speed`, `engine_rpm`, `gear` and chassis `drag`.

Those are precisely the signals the XIL dyno loop needs and 0.9.15 does not expose.
On 0.9.15 the wheel speed and tire torque have to be reconstructed from the vehicle's
longitudinal equation; on 0.9.16 they can be read directly, per wheel. See
`doc/xil/CarlaDynoCoupling.md` for what that changes.

This issue is **investigation only**. No 0.9.16 change should land on a major branch
until we decide to migrate.

## What has been verified

0.9.16 was built from source on Windows (`C:\src_ext\CarlaSrc_0_9_16`, tag `0.9.16`)
against the **same UE 4.26 fork** already in use — no engine rebuild required.

- `make setup` — clean, after fixes below
- `make PythonAPI` — produced `carla-0.9.16-cp310-cp310-win_amd64.whl`
- telemetry API confirmed live in the built client: `get_telemetry_data`,
  `WheelTelemetryData` with all 11 fields, `VehicleTelemetryData`
- `make package` — in progress at time of writing

## Build fixes required (not needed for 0.9.15)

| # | Problem | Fix |
|---|---|---|
| 1 | `Setup.bat` hardcodes `--boost-toolset msvc-14.3`; our 0.9.15 build used `vc142` (VS2019) | build with **VS2022**; generator `Visual Studio 17 2022` |
| 2 | `cd "%VAR%"` where the variable ends in `\` — the `\"` escapes the quote, so `cd` fails silently | append a trailing dot: `cd "%VAR%."` — 7 scripts in `Util/InstallersWin` |
| 3 | `install_boost.bat` calls `bootstrap.bat vc141` (VS2017), which is not installed | change to `vc143` |
| 4 | `python -m build` used for the wheel; the `build` module was absent | `pip install build distro` for Python 3.10 |
| 5 | `build`'s isolated env fails (pip 23.0.1 vs setuptools 84) | add `--no-isolation` in `BuildPythonAPI.bat` |
[...]

PR #323, opened 2026-08-26
#323/a simulated XIL dyno bench in CommonLib, and the CARLA coupling studies on it

yunlishao-vibe:
## Summary

A simulated XIL dyno bench that ships as FIXS infrastructure, and the coupling
studies built on it. Nothing here imports CARLA, FIXS runtime or numpy — standard
library only, so it loads and unit-tests on a machine with neither simulator.

### `CommonLib/xil` — the parts of the setup, and the simulation that couples them

```
v_ref ─▶ [RobotDriver] ─▶ pedals ─▶ [DynoSim: Vehicle on Dyno] ─▶ v_measured
         who works                   the car, the machine, and
         the pedal                   the equations of motion

... and [link] carries v_ref out and v_measured back, in one process or over
    UDP in the 8-byte 2xfloat32 format the ORNL cell speaks on 5010/5011
```

```
vehicle.py   Vehicle, Powertrain, Driveline    the car under test
driver.py    RobotDriver                       who works the pedal
link.py      LocalLink, UdpLink                the wire out to CARLA
dynosim.py   Dyno, DynoSim, State              the dyno, and a car on it
```

Three modules are parts you can point at in a test cell. `dynosim.py` is the dyno
side: the machine itself, and the equations of motion that only exist once a car
is bolted to it — the acceleration of a car on a dyno belongs to neither the car
nor the dyno alone, and the effective mass it turns on is the car plus the roller
plus four wheels referred to the road through the tyre radius.

`mode='chassis'` puts the vehicle on rollers; `mode='axle'` replaces the wheels
with hub units and emulates the body's inertia electrically, since there is none.

`__init__.py` re-exports nothing on purpose — import from the module that defines
the thing, so the import line says where it lives.

```python
from CommonLib.xil.dynosim import Dyno, DynoSim
from CommonLib.xil.vehicle import Vehicle

sim = DynoSim(Vehicle(mass_kg=2100), Dyno(mode='axle'))
state = sim.step(v_ref, dt)               # the driver chases a reference
state = sim.step_pedals(thr, brk, dt)     # or work the pedal yourself
```

**The dyno applies road resistance and nothing else.** An earlier revision also
offered a speed-controlled mode where the dyno servo forced a commanded speed.
That is real hardware, but it does not compose with a speed-tracking driver —
both steer the same degree of freedom, so the driver's error sits at zero and its
pedal parks on whatever its integrator held. Measured: reaching 15 m/s three
different ways parked the throttle at 0.040, 0.206 and 0.696, and the "measured"
force varied by the same 17x. Under road resistance the pedal is pinned by
physics: 0.0283 every time, because that is the value whose tractive force
balances it.

**The pedal is always an input.** A real bench has a real vehicle on it whose
pedal somebody works. `RobotDriver` stands in for that, and its integral term is
not optional — a steady speed needs a steady pedal, and a proportional-only law
can only produce one from a standing error.

### Three defects the build surfaced

**An unstable servo shipped in the first draft.** `kp = 900` against a 25 Hz
actuator is 62 Hz crossover and **-47 degrees of phase margin**; the bench hunted
±8 rad/s instead of holding speed. Gains are now derived from a requested
bandwidth and the inertia they act on, because for `I·dx/dt = u` a PI crosses
over at `kp/I` and a gain quoted without its inertia is meaningless.

**`front_share` was destroying half the vehicle.** It scaled each axle's
*capability* by its share instead of splitting the *demand*, so full throttle
delivered 3280 of 6400 Nm — 127 kW of a rated 239 — and terminal speed was wrong
for a reason that had nothing to do with the road. Now 99.8 % at full throttle,
share honoured exactly below the caps, and `front_share = 0` correctly means
rear-drive only.

**Road resistance accelerated a parked car backwards.** It was summed as a signed
term. Applied and resistive efforts are now kept apart, resistive ones only
opposing the motion that exists. Verified: exact zero held for 200 s with no
creep, breaking away at the throttle the algebra predicts.

Also: the bench **sub-steps internally**. Handed 0.1 s — the TrafficLayer feed
period — it diverged to 1e81 and the clamps then snapped it to zero, which reads
as a stopped vehicle rather than a broken integration. Identical answers now from
1 ms to 200 ms.

### `tests/Xil/Probes/dyno_speed_sync/` — two studies

`dyno_sync_sim.py` prices four ways of coupling CARLA to the bench. The
disagreement between the two resistance curves **changes sign at ~20 m/s**, so no
constant trim removes it; syncing faster buys accuracy and not quietness, because
`F_sync = m·dv/T` with `dv` growing in `T`; and whether CARLA drives itself
dominates everything — coasting makes the overwrite supply the whole tractive
effort at 2682 N rms against 867 N when it does not.

`tracking_scenarios.py` asks the narrower question — **how should CARLA be told to
follow the bench?** — over free driving and a braking leader, and writes a plotly
page. The two options are the two things a controller can write through
`ego.set`, the same choice `COMMAND_SHAPE` makes in `ego_agent_controller.py`:

```
  scenario command   err_rms  err_mean  x_diverge[m]  T_bench_pk  T_carla_pk  ratio  %capac
  free     speed      0.0173   -0.0004         -0.03        2169        1712   0.79     58%
  free     pedals     0.3770   -0.0213         -1.92        2169        2349   1.08     57%
  leader   speed      0.0095    0.0006          0.05        1491        1191   0.80     21%
  leader   pedals     0.1216   -0.0010         -0.09        1491        1422   0.95     22%
```

**The torque columns are the point.** Tracking well is not tracking honestly. The
two plants are 33 % apart in effective inertia (2452 kg against 1845) and their
resistance curves cross, so for CARLA to hold the bench's speed it must produce a
*different* force — and a stiff loop will deliver that difference without
complaint while the speed trace looks perfect. Under a speed command CARLA peaks
at **0.79x** the bench's torque, which is what the mass ratio predicts
(`1845/2452 = 0.75`), and never exceeds **58 %** of its capability. Had it come
back at 3x, the speed command would be manufacturing a correct-looking trajectory
out of a fictional vehicle.

`kp = 0.55` and the `a/3.2` pedal map come from ORNL's
`carla_standalone_drive.py`. The integral term does **not** — theirs is
proportional only, which leaves CARLA 0.44 m/s slow and 40 m behind over 90 s.
`PEDAL_KI = 0` restores their law.

## Verification

```
pytest tests/Python/unit/test_xil_dyno.py tests/Python/unit/test_xil_link.py -q
pytest tests/Xil/Probes/dyno_speed_sync/ -q            55 passed together
pytest tests/Python/unit/ -q                          276 passed, 8 skipped
```

Rebased onto current `dev_v0.9.0`.

## Related Issues

Companion to `FIXS_Applications#24`. Does not close anything. Adjacent and not
blocking: #322 (per-axle torque injection) and #319 (0.9.16 migration).

## Environment
- Python version: 3.10.20 (`realsim`); plotly for the tracking figure,
  matplotlib optional for the coupling figure

## Checklist
- [x] Tests pass locally
- [x] Issue linked above

## Additional Notes

Two things a reviewer should know.

The CARLA side of both studies is a **surrogate** — a longitudinal ODE
parameterised from constants measured off a live 0.9.16 server, not CARLA. It
exists so the studies run without a server. Hooking the real thing is pointing
the link away from it.

And the weakest input is CARLA's `roll_coeff = 0.012`, which is **not measured**:
it lives inside CARLA's tire model and only surfaces lumped into
`sum(long_force)`. It decides where the resistance gap changes sign, which is the
whole character of the mismatch, so it is the first thing to replace once a
server-connected run exists.


yunlishao-vibe:
Re-ran everything under the `realsim` conda env (3.10.20): **13 tests pass**, and the study numbers are identical to those in the PR body. Also identical under `fixs_applications`, which is the one to use if you want the figure since it has matplotlib. The simulator itself needs only the standard library.

README updated to name `realsim` rather than `realsim_dev`.

yunlishao-vibe:
Speed tracking is now a control mode, orthogonal to where the dyno couples — `mode` picks roller vs hub units, `control` picks road-load emulation vs speed servo. All four combinations work and are tested.

`control='speed'` is the one for coupling to a simulator that already owns the vehicle dynamics: it holds the commanded speed and the measurement is the effort that took. It deliberately does not impose road load, since the servo has replaced the road and double-counting it would be silent.

Verified on chassis+speed: at throttle 0.3 and 15 m/s the envelope delivers 867.6 Nm, which over r = 0.36 m is 2410 N, and the dyno absorbs −2409.9 N with net zero on the vehicle.

33 tests on the bench, 12 on the coupling, full unit suite 267 passed.

yunlishao-vibe:
The bench is now three pieces, matching the three the real bench has, and the dyno does one thing.

**Road resistance only.** The speed-controlled dyno mode is gone. It does not compose with a speed-tracking driver — both steer the same degree of freedom, so the driver's error sits at zero and its pedal parks on whatever its integrator held. Reaching 15 m/s three different ways parked the throttle at 0.040, 0.206 and 0.696, and the measured force varied by the same 17x. Under road resistance the pedal is pinned by physics: 0.0283 every time.

**`RobotDriver`** is the pedal authority — PI on `v_ref − v_measured`, integrator not optional. **`link`** carries `v_ref` out and `v_measured` back, in-process or over UDP in the 8-byte 2×float32 format the ORNL cell speaks on 5010/5011. Steer is relayed but nothing acts on it. `DynoVehicle` composes driver and bench so the pair presents the interface the wire carries, and the study drives that over the link — so pointing at a real CARLA later is a transport change, not a code change.

**Two bugs, both found by disbelieving an output.** `front_share` was scaling each axle's capability rather than splitting the demand, so full throttle delivered 51% of the vehicle (3280 of 6400 Nm, 127 of 239 kW). And there was no speed limiter, so the bench settled at 229 km/h where an EV6 stops at 185.

**The bench now sub-steps internally.** Handed 0.1 s — the TrafficLayer feed period — it diverged to 1e81 and the clamps snapped it to zero, which reads as a stopped vehicle rather than a broken integration. Verified identical answers from 1 ms to 200 ms.

46 tests on the bench and wire, 13 on the coupling, full unit suite 280 passed.

Worth knowing: this machine crashes roughly two runs in three with a Windows fatal exception unrelated to any assertion. The same code passes when it completes and no crash has reproduced under isolation.

yunlishao (approved):

Issue #327, opened 2026-08-27
[FEATURE] sumo_uturn: attach drivable U-turn loops to any SUMO corridor end


yunlishao:
**Is your feature request related to a problem?**

A corridor scenario that stops at a stub has no way to send a vehicle back the other way except SUMO's *point* turnaround: a short internal link that pivots the vehicle 180 deg on the spot. On the MLK network those links are 4.97 m (east, 3.73 m/s) and 10.14 m (west, 4.89 m/s) — a turning radius of roughly 1.6 m. That is fine for queue statistics and unusable for anything that has to physically drive the geometry: a CARLA/CarMaker ego following the OpenDRIVE export cannot track it, so the ego either halts at the terminus or leaves the road.

FIXS already solved this once. For #172 `gen_loop_net.py` builds the SimpleTrafficLight corridor with a compact curved turnaround ("teardrop") at each end — entry ramp, ~300 deg loop, exit ramp, all at a 45 m radius. But that script *generates a whole synthetic corridor*; there is no way to put the same geometry on a network that already exists, which is what every real digital twin needs.

**Describe the solution you'd like**

`Carla/utils/sumo_uturn.py` — the geometry core of `gen_loop_net.py`, generalised into a tool that patches an existing `.net.xml`:

- works at any heading and carries the terminus elevation, so it applies to a real network, not just an axis-aligned synthetic one;
- keeps every existing edge id (it adds edges, never renames or removes), so route files stay valid;
- finds corridor termini by geometry — a junction whose only through movement turns back on itself — with `--list`, or takes them from an ordered corridor edge CSV, a junction id, or an explicit edge pair;
- attaches to the *lane centrelines* that carry the traffic rather than the junction centroid, and blends the stub with a cubic Hermite, so no kink is introduced at the join;
- emits plain-XML node/edge/connection patches and hands them to `netconvert`, so nothing is hand-edited in the net;
- removes the point turnaround it replaces (via `<delete>`; netconvert regenerates turnarounds otherwise) and reports how many existing routes that breaks.

`gen_loop_net.py` should then import the shared core instead of keeping its own copy of the curvature integration, so the two cannot drift.

**Describe alternatives you've considered**

- *Hand-editing the net.xml.* Fragile: removing a connection means fixing the junction's `intLanes`, `request`/`response`/`foes` indices by hand.
[...]

PR #332, opened 2026-08-27
run_cosim: --sumo-only, for a scenario CARLA cannot render yet (#331)

### LINKED ISSUE #331: [FEATURE] run_cosim --sumo-only: drive a scenario CARLA cannot render yet

`run_cosim` always launches VirCarlaEnv — it is the CARLA co-sim front door — and every scenario yaml carries a `CarlaSetup` section, because that is what makes it a co-sim yaml. There is no way to drive a scenario whose SUMO geometry CARLA has no counterpart for.

That is not hypothetical. The MLK corridor's ends were just given drivable U-turn loops (#327): ~390 m of road per terminus that the CARLA export does not contain. Every config in that app's picker would place the ego on road CARLA has never heard of, for ~45 s per turnaround, twice a lap — silently, looking like a bridge or coordinate-offset bug. The app's own guard now refuses those runs, which leaves the network correct and unrunnable through the front door.

The same gap applies to any network edited on the SUMO side, and to a controller change worth checking before anyone cooks a map for it.

`--sumo-only`: drop VirCarlaEnv, change nothing else — same SUMO, same TrafficLayer, same app controller, same ports, same teardown. Return before the CARLA preflight rather than threading a flag through it; that block exists to put a specific map in front of a specific server, and a run with no server has no part of it to keep.

Two things fall out:

- **Something has to be blocked on.** The supervision loop blocks on VirCarlaEnv today. Generalise it to a `front` process: the app's controller where there is one, TrafficLayer where there is not — keeping the existing "report whoever died first" behaviour rather than adding a second loop.
- **The app has to be told.** An app cannot infer "is CARLA in this run" from the yaml, since every co-sim yaml has a `CarlaSetup` section either way. An app that refuses to run geometry CARLA cannot render would then refuse the very run that exists to drive it without CARLA. So export `FIXS_SUMO_ONLY=1` alongside the existing `FIXS_HANDOFF` / `FIXS_PYTHON` / `FIXS_CONFIG_YAML` contract.

**Hidden from `--help`** (`argparse.SUPPRESS`) for now. run_cosim's promise is that what you pick is a co-simulation; advertising "...but without the simulator half" invites picking it to make a CARLA problem go away, which is how you get results nobody can place.

- *A SUMO-only config yaml with no `CarlaSetup`.* Does not help — run_cosim launches VirCarlaEnv regardless, and the picker only offers what `apps.json` lists.
- *A per-app SUMO-only launcher.* Works (MLK has one for sumo-gui alone), but it runs neither TrafficLayer nor the controller, so it cannot check a controller change, and every app would need its own.

**Environment**
- Python version: 3.10
- SUMO version: 1.21
- CARLA version: n/a — the point is to not need one

**Additional context**

Fully verified only as far as it can be without a build: compiles, hidden from `--help`, flag parses, every VirCarlaEnv reference confined to the non-sumo_only branch, `front` bound on both paths, and `peer.gone(None)` already returns `False` so the supervision loop is safe with no control socket. An end-to-end run needs a FIXS build that contains it — the installed v0.9.0-alpha bundle predates it.

yunlishao-vibe:
Closes #331.

## Why

`run_cosim` always launches VirCarlaEnv — it is the CARLA co-sim front door — and every scenario yaml carries a `CarlaSetup` section, because that is what makes it a co-sim yaml. So there was no way to drive a scenario whose SUMO geometry CARLA has no counterpart for.

Concretely: MLK's corridor ends just gained drivable U-turn loops (#327), ~390 m of road per terminus that the CARLA export does not contain. Every config in that app's picker would put the ego on road CARLA has never heard of, for ~45 s per turnaround, twice a lap. The app's guard now refuses those runs — which left the network correct and unrunnable through the front door.

## What

`--sumo-only` drops VirCarlaEnv and changes nothing else: same SUMO, same TrafficLayer, same app controller, same ports, same teardown.

It **returns before the CARLA preflight** rather than threading a flag through it. That block exists to put a specific map in front of a specific server; a run with no server has no part of it to keep, and gating it in place would have re-indented ~160 lines to say "skip all of this".

Two consequences worth calling out:

**Something has to be blocked on.** The supervision loop blocked on VirCarlaEnv. It now blocks on a `front` process — the app's controller where there is one, TrafficLayer where there is not — keeping the existing "report whoever died first" behaviour rather than growing a second loop.

**The app has to be told.** An app cannot infer *is CARLA in this run* from the yaml, since every co-sim yaml has a `CarlaSetup` section either way. An app that refuses to run geometry CARLA cannot render would then refuse the very run that exists to drive it without CARLA. So `FIXS_SUMO_ONLY=1` joins the `FIXS_HANDOFF` / `FIXS_PYTHON` / `FIXS_CONFIG_YAML` contract — a fourth variable, added for the same reason as the other three: only run_cosim knows, so run_cosim says.

## Hidden from `--help`

`argparse.SUPPRESS`. run_cosim's promise is that what you pick is a co-simulation, and advertising "...but without the simulator half" invites picking it to make a CARLA problem go away — which is how you get results nobody can place. The comment says to un-suppress it when it is something users are meant to reach for.

## Verification, and its limit

Verified: compiles; `--help` does not mention it; the flag parses; every VirCarlaEnv reference in `run_native_stack` sits inside the non-`sumo_only` branch and `front` is bound on both paths; `peer.gone(None)` already returns `False`, so the supervision loop is safe with no control socket.

**Not verified end to end.** The installed v0.9.0-alpha bundle predates this change (it is built from a tree ahead of `dev_v0.9.0` — it carries `fix/cosim_setup_gaps`), so I could not run a real stack through it without patching a released bundle in place. What settles it: a FIXS build containing this, then `run_cosim.bat --sumo-only` on `mlk_eco_driving` with the U-turn network — expected `[cosim] SUMO-only stack up: SUMO + TrafficLayer + mlk_eco_driving, no CARLA.` and the ego lapping both loops.

### DIFF
--- a/Carla/run_cosim.py
+++ b/Carla/run_cosim.py
@@ -722,7 +722,8 @@ CarlaSetup:
 HANDOFF_TIMEOUT_S = 60
 
 
-def start_app(app, config_yaml=None, timeout=HANDOFF_TIMEOUT_S, sumocfg=None):
+def start_app(app, config_yaml=None, timeout=HANDOFF_TIMEOUT_S, sumo_only=False,
+              sumocfg=None):
     """Start the app's `launch` command and collect the scenario it reports.
 
     Returns (proc, sumocfg-or-None). ONE process spans both moments a controller
@@ -761,6 +762,14 @@ def start_app(app, config_yaml=None, timeout=HANDOFF_TIMEOUT_S, sumocfg=None):
     env = dict(os.environ, FIXS_HANDOFF=handoff, FIXS_PYTHON=sys.executable)
     if config_yaml:
         env["FIXS_CONFIG_YAML"] = config_yaml
+    # FIXS_SUMO_ONLY because the yaml cannot answer "is CARLA in this run". Every
+    # co-sim yaml has a CarlaSetup section - that is what makes it a co-sim yaml -
+    # so an app checking the yaml to decide whether its geometry will be rendered
+    # gets the same answer either way, and one that refuses to run a scenario CARLA
+    # cannot render would refuse the very run that exists to do it without CARLA.
+    # Only run_cosim knows, so run_cosim says.
+    if sumo_only:
+        env["FIXS_SUMO_ONLY"] = "1"
     # FIXS_SUMOCFG because --sumocfg names the scenario THE USER chose, and an app
     # that generates its own can only honour that by building FROM it. Without this
     # the flag reaches SUMO but never the app, so the only thing it could do to a
@@ -800,7 +809,7 @@ def start_app(app, config_yaml=None, timeout=HANDOFF_TIMEOUT_S, sumocfg=None):
 
 def run_native_stack(config_yaml, sumocfg, tl_table, cfg, args, app=None,
                      ctl_sock=None, app_owns_scenario=False, app_proc=None,
-                     bridge="cpp"):
+                      sumo_only=False, bridge="cpp"):
     """FIXS-native bridge: launch SUMO (TraCI server) + TrafficLayer (-f config) +
     VirCarlaEnv (-f config -t tl_table), plus the app's own `launch` command if it
     declares one. CARLA is already up and the map loaded by run_cosim's preflight.
@@ -809,13 +818,21 @@ def run_native_stack(config_yaml, sumocfg, tl_table, cfg, args, app=None,
     TrafficLayer is the sole TraCI client, so it steps SUMO (an app controller
     subscribes to TrafficLayer on the application port, it does not drive SUMO).
     Blocks on VirCarlaEnv - the co-sim front-end - and tears the others down on
-    exit."""
+    exit.
+
+    sumo_only drops VirCarlaEnv and changes nothing else: same SUMO, same
+    TrafficLayer, same app controller, same ports. It is for driving a scenario
+    CARLA cannot render - a network edited on the SUMO side, or a controller
+    change worth checking before anyone cooks a map for it. With no CARLA
+    front-end there is nothing to block on, so it blocks on the app's controller
+    where there is one and on TrafficLayer where there is not."""
     import shutil
     tl_exe, vce_exe = _native_binaries()
-    # Only require the bridge this run will actually start. The C++ VirCarlaEnv is
-    # built only when the libcarla dep is available, so demanding it for a Python-
-    # bridge run would refuse a stack that has everything it needs.
-    needed = [tl_exe] + ([vce_exe] if bridge == "cpp" else [PY_BRIDGE])
+    # Only require what this run will actually start: the bridge it selected, and no
+    # bridge at all when there is no CARLA half. The C++ VirCarlaEnv is built only
+    # when the libcarla dep is available, so demanding it for a Python-bridge run -
+    # or for a SUMO-only one - would refuse a stack that has everything it needs.
+    needed = [tl_exe] if sumo_only else [tl_exe, vce_exe if bridge == "cpp" else PY_BRIDGE]
     missing = [q for q in needed if not os.path.isfile(q)]
     if missing:
         sys.exit(f"[cosim] engine {bridge!r} needs these, not found:" + "\n  "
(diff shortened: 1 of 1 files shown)


yunlishao-vibe:
### Now verified end to end

The PR said this was unverified because the installed v0.9.0-alpha bundle predates it. I hot-landed the change onto a copy of that bundle instead — the bundle's own `run_native_stack` is ahead of `dev_v0.9.0` (it already generalises the front process as `bridge_name` for the two-bridge work), so the same semantic change was applied against the bundle's text rather than this diff. Every edit anchored and asserted, so a missing anchor stops rather than half-patches.

`run_cosim.bat --sumo-only` on `mlk_eco_driving`, U-turn network:

```
[cosim] --sumo-only: SUMO + TrafficLayer + mlk_eco_driving, no CARLA. config ...config_Sumo_Carla_ecoDriving.yaml
[cosim] cadence (cpp): FIXS feed 0.1 s ... pacing as fast as possible
[SUMO] sumo-gui -c ...\mlk_simulation_ego.sumocfg --remote-port 1337 ...
[cosim]   OK   SUMO running (pid 21436)
[cosim]   OK   SUMO TraCI listening on 1337
[TL]   TrafficLayer.exe -f ...config_Sumo_Carla_ecoDriving.yaml
[cosim]   OK   TrafficLayer running (pid 65644)
[cosim]   OK   TrafficLayer serving the bridge on 440

[cosim] SUMO-only stack up: SUMO + TrafficLayer + mlk_eco_driving, no CARLA.
[cosim] nothing is rendered - watch it in sumo-gui. Ctrl+C here, or let mlk_eco_driving finish, to stop.
```

No VirCarlaEnv process, no CARLA connect, no world load. Note the config is `config_Sumo_Carla_ecoDriving.yaml` — a yaml with a `CarlaSetup` section — which is the case that motivated `FIXS_SUMO_ONLY`: the app's geometry guard let it through on the env var, having refused the same yaml without it.

### One thing it caught

The first attempt died in the app, not here:

```
FileNotFoundError: ...\prepared_scenario\MLK_final_elevation_20251009.net.xml
[cosim] 'mlk_eco_driving' exited (1) before reporting its scenario; not starting the stack.
```

The MLK controller kept its own hardcoded default for the network name — a third owner of it, behind the sumocfg and the `.env`. Fixed app-side (ORNL-Real-Sim/FIXS_Applications#47); nothing to change here. Worth noting only because it is the failure mode `FIXS_SUMO_ONLY` is meant to make *visible*: run_cosim reported which component exited and did not start a stack around it.

yunlishao-vibe:
### Correction to the comment above

I said "verified end to end" too early. The stack coming up is not a run completing, and I posted before letting one finish. What the runs actually show, and the two bugs they found, are in `fac0fa71`.

**What is verified.** With a scenario yaml that has no bridge subscriber (MLK's `config_Sumo_ecoDriving.yaml`, port 430 only):

```
[cosim] --sumo-only: SUMO + TrafficLayer + mlk_eco_driving, no CARLA.
Client ports: 430
All Clients Connected!
[cosim] SUMO-only stack up: SUMO + TrafficLayer + mlk_eco_driving, no CARLA.
FIXS simulation time: 28800.0996 ... 29420.9004
```

620 s of simulation, the eco controller genuinely in the loop (computing an eco speed against a red at `202695971`, ego on `pos_78507374#0`), SUMO stepped by TrafficLayer, no VirCarlaEnv process and no CARLA connection anywhere. That is the feature working.

**What is not.** TrafficLayer then exited `4294967295` at t≈29420 — about 60 s of sim time before the ego's first U-turn at ~29481 — and the controller died on the resulting connection reset. Its own log has nothing beyond the config line. I have not diagnosed it and I am not claiming it is unrelated to this change. The decisive test is the same run on the stock network: if TrafficLayer exits there too, it is independent of `--sumo-only` and of the U-turn geometry.

**Two bugs the runs found, both fixed here:**

1. The `backend != "cpp"` check was wrong. `backend` picks *which* CARLA bridge would run (`cpp` = VirCarlaEnv, `py` = run_synchronization); `--sumo-only` runs neither, so refusing `py` rejected a SUMO-only scenario for naming a bridge nothing would start.

2. A CARLA yaml lists the bridge port among TrafficLayer's subscribers. With no bridge, TrafficLayer waits for a client that never connects and tears the run down at the end of warm-up — the app then dies on `WinError 10054`, which reads as an app bug with the real cause several steps upstream and mentioned nowhere. Observed exactly that with `config_Sumo_Carla_ecoDriving.yaml` (subscriber on 440): warm-up ran 28800 → 29010, then the reset. `--sumo-only` now warns up front and names what to use instead.

The second one is worth a design opinion from a reviewer: a warning is the conservative choice, but the failure it predicts is fatal every time, so it may deserve to be an error.

yunlishao (approved):

PR #338, opened 2026-09-01
cosim: --purge-map, to remove imported maps from a machine

yunlishao-vibe:
## Why

An imported map lands in **three** places that age independently, and nothing removed any of them:

| location | rebuilt by | read at run time? |
|---|---|---|
| `Content/<name>/` | a re-cook | no |
| `Import/<name>/` + `<name>.json` | a re-stage | no |
| `~/.fixs/maps/<name>/` | a **download** | **yes** — `bundle_sumocfg` reads the scenario out of its `sumo/` |

A cook that crashed left content behind that no re-import clears, and reclaiming the disk meant knowing all three paths by heart.

## What

`--purge-map` lists what every map occupies, in those three columns, and deletes the ones chosen. It stops there; it is not a run.

```
[purge] maps on this machine (5.5 GB in total):
     #  map                       cooked    staged     cache
     1) atlanta_full            108.7 MB  300.8 MB  316.6 MB
     2) mlk_no_signal            52.8 MB  108.6 MB   53.3 MB
     3) mlk_untextured           86.4 MB   45.6 MB         -
     4) Roosevelt06082026              -  733.8 MB         -   (staging only, never cooked)

[purge] Purge which? [1-4, a list like 1,3,4, or 'all'; Enter to cancel]:
```

- pick by number (`1,3,4`, a `2-5` range, or `all`), or by name: `--purge-map mlk_untextured,atlanta_full`, `--purge-map all`
- every path and size is printed and confirmed before anything is deleted

**The cache is chosen separately rather than swept along.** It is the only half that costs a download to rebuild, and it is also the half a *run* reads — so dropping it does not merely slow the next import, it stops the next run until something re-fetches it. Different blast radius, its own yes. `--purge-cache` / `--keep-cache` answer it up front for scripts.

## Two traps, both covered by tests

- **CARLA's own `Content/Carla/` ships a `Config/Carla.Package.json`**, exactly like an imported package does — so "has a package descriptor" would offer the engine's own content up for deletion. Qualification is on the map's own `.umap`, with an explicit engine deny-list behind it.
- **`_map_cache_dir(name)` creates the folder it names.** Using it to take an inventory invents an empty cache for every map and then reports it as purgeable. `purge_candidates` joins the path itself.

Also handled: a cook that died part-way (descriptor, no `.umap`) is listed and flagged, since that wreckage is most of the reason to want this; the `.bak_reimport` a failed re-import strands is swept with its map; a lone `Import/<name>.json` never qualifies a name on its own, because CARLA's `Import/` holds descriptors like `roadpainter_decals.json` that name no map; `carla_root=None` (client mode, or no CARLA configured) still reports the bundle cache.

Refuses while CARLA is running rather than half-deleting a package Windows will not unlink. Dispatched beside `--doctor` — it needs the saved config to find CARLA but never the `carla` module, so it must not pay for the re-exec.

## Front door

`run_cosim.bat` forwards unknown flags untouched, so `run_cosim.bat --purge-map` works with no change there. Adding it to that script's `--help` waits for a FIXS release carrying this, so the front door never advertises a flag the installed engine rejects.

## Testing

27 new Tier-1 tests (pure stdlib, fake CARLA trees, no server/GPU/map). `253 passed, 8 skipped` across `tests/Sumo/Carla` + `tests/Python`. End-to-end flow driven against a sandbox tree: engine content never offered, only the picked maps removed, the unpicked map and its cache untouched.

Diff is +540 / -0.

### DIFF
--- a/Carla/import_map.py
+++ b/Carla/import_map.py
@@ -2362,6 +2362,191 @@ def purge_maps(records, drop_cache=False):
     return removed, failed
 
 
+def _purge_ask(msg):
+    """input() where EOF means 'answered nothing'. Deliberately not _prompt: that
+    one exits with advice about --map / --package, which is the wrong advice here
+    and, worse, exits mid-command. A purge that cannot ask must cancel, never
+    fall through to a default - the next step deletes."""
+    try:
+        return input(msg).strip()
+    except EOFError:
+        return ""
+
+
+def _parse_selection(answer, count):
+    """0-based indices named by a picker answer: '2', '1,3,4', '2-5', 'all'. None
+    when the answer does not parse or names nothing - the caller re-asks or
+    cancels rather than acting on a guess, because what follows is a deletion."""
+    answer = (answer or "").strip().lower()
+    if answer in ("a", "all", "*"):
+        return list(range(count))
+    chosen = set()
+    for part in answer.replace(" ", "").split(","):
+        if not part:
+            continue
+        lo, dash, hi = part.partition("-")
+        if dash:
+            if not (lo.isdigit() and hi.isdigit()):
+                return None
+            lo, hi = int(lo), int(hi)
+            if not (1 <= lo <= hi <= count):
+                return None
+            chosen.update(range(lo - 1, hi))
+        elif part.isdigit() and 1 <= int(part) <= count:
+            chosen.add(int(part) - 1)
+        else:
+            return None
+    return sorted(chosen) or None
+
+
+def _purge_table(records, indent="   "):
+    """The inventory as a numbered table: one row per map, one column per place a
+    map occupies. Three columns rather than one total, because they are not
+    equally expensive to rebuild and whoever is choosing needs to see that before
+    they choose."""
+    print(f"{indent}{'#':>3}  {'map':<22}{'cooked':>10}{'staged':>10}{'cache':>10}")
+    for i, record in enumerate(records, 1):
+        by = {}
+        for label, _path, size in record["pieces"]:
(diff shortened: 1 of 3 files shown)


yunlishao-vibe:
Restructured after review: the implementation moved from `run_cosim.py` into `import_map.py`.

`--doctor` already answers this question in this codebase — a front-door action is a thin dispatch in `run_cosim.py`, the implementation lives in the module that owns the domain, and what only `run_cosim` can answer is injected as a parameter (`doctor.run(..., who_has_port=_who_has_port)`). The first cut ignored that and put 191 lines in `run_cosim.py`.

`purge()`, the picker and the table now sit in `import_map.py` beside the path rules they depend on — `content_root`, `cooked_map_path`, `package_descriptor`, `_map_cache_dir` — several private, all of which resolve differently per CARLA flavour. One home for those rules again, rather than a second reader that could go stale inside the code that deletes things.

`run_cosim.py` keeps the three flags plus a dispatch, and passes in the two things only it can answer: whether anyone is there to be asked (`_interactive` — a `--serve` host has a terminal and nobody at it, which `isatty` cannot tell), and whether a CARLA is holding `Content/` open.

**No behaviour change** — same flags, output and guards. `run_cosim.py` is now **+41** instead of +191.

Evidence the seam is right: the end-to-end smoke used to monkeypatch `run_cosim._interactive`, `_ask` and `_pid_on_port` to drive the flow. It now just calls `import_map.purge(...)` and passes those in.

`253 passed, 8 skipped`.

yunlishao (approved):

PR #343, opened 2026-09-03
cosim: --reimport can take a path, and stop asking a question with one answer

yunlishao-vibe:
A local map's saved setup remembers exactly one thing about its source: a
folder path. `--reimport` used to trust that silently, forever - which goes
stale the moment the export moves (the next `--reimport` either errors on a
folder that no longer exists, or worse, still finds *something* there and
silently re-cooks old files, with no error at all).

`--reimport` now parses as an optional value, the same shape `--purge-map`
already uses:

```
run_cosim.bat --profile uga_dev --reimport                    (reopens the file browser)
run_cosim.bat --profile uga_dev --reimport "G:\...\UGA_Exports"   (skips it, uses this path)
```

Either way the result becomes this run's source, and the existing pre-cook
checkpoint save already writes it back into the setup - so whatever was just
chosen is what the *next* `--reimport` remembers. No new persistence code;
that plumbing already existed for a different reason.

A path only means something for a local pick - a Digital-Twin-Library map's
`--reimport` is a re-download, not a re-browse:

```
[cosim] --reimport 'G:\...\export' ignored: 'roosevelt_full' is a
Digital-Twin-Library map, not a local pick.
```

**And this is what finally lets a redundant staging prompt (added in #342) go
quiet where it should.** `resolve_existing_staging`'s `have_source=True` branch was
always the "restage from what was just given" case with only one possible
answer - and now that a source is *always* supplied (a fresh browse or an
explicit path), asking about it stopped every single `--reimport` for an
Enter key confirming a certainty:

```
[import] replacing the staged package for 'uga' with the source you gave.
```

No question. The interactive `[U]se/[A]bort` gate is kept, unchanged, for the
one case it actually protects: replaying a saved setup with neither a fresh
pick nor `--reimport` at all, where FIXS genuinely doesn't know where
`Import/<name>/`'s contents came from.

Verified: the three-way `--reimport` parse (absent / bare / path), the silent
restage on `have_source=True`, the unchanged prompt on `have_source=False`,
and the DT-Library ignore-and-report path - against the real CARLA install.

### DIFF
--- a/Carla/import_map.py
+++ b/Carla/import_map.py
@@ -2527,45 +2527,42 @@ def resolve_existing_staging(carla_root, name, have_source, interactive=None):
     """Report an Import/ staging that already exists for `name`, and decide what to
     do about it. Returns "use", "restage" or "abort".
 
-    This is deliberately loud. Import/<name>/ is the input CARLA cooks, so a copy
-    left by an earlier import of a DIFFERENT export under the same name is cooked
-    in silence - the map builds, it is simply not the map that was picked. The old
-    behaviour said one quiet "package already staged" line for exactly that case.
+    Only INTERRUPTS when have_source is False - when nothing tells this call where
+    to copy from, so reusing whatever is already in Import/<name>/ is a guess: it
+    could be a leftover from an earlier import of a DIFFERENT export that happened
+    to share this name. That guess is what deserves a human's eyes.
 
-    The default is what this function would have done before it existed - reuse
-    when nothing else was supplied, re-stage when a source was - so pressing Enter
-    changes nothing. Non-interactive keeps that default and still prints the
-    report, which is the part that has to survive into the log."""
+    have_source=True means the opposite: a path, a URL, or an explicit pick was
+    already given, so there is nothing to decide - restaging from it is not a
+    guess, it is what was just asked for. Interrupting there anyway (every prior
+    version of this function did) turned a certainty into a keystroke on every
+    single --reimport, forever, for a caller who by definition already knows the
+    answer before being asked."""
     existing = staging_report(carla_root, name)
     if not existing:
         return "restage" if have_source else "use"
 
-    default = "restage" if have_source else "use"
+    if have_source:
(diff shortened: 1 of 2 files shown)


yunlishao (approved):

yunlishao-vibe:
---

## Follow-up: two gaps found by auditing this fix

Testing `--reimport PATH` reproduced the exact bug it was meant to solve:
CARLA asked first, cooked, *then* SUMO asked mid-cook. Auditing every place a
local map's CARLA source can be freshly established found two bugs - one
older than this PR, one introduced by the commit above.

**Bug 1 (pre-existing).** `_sumo_for_local_pick`'s guard read
`if args.sumocfg is not None or args.reimport: return` - skipping the early
ask whenever `--reimport` was set. The reasoning ("the late chain handles it
anyway") missed that avoiding the *late* ask was the entire point. Root
cause: `cached_sumo_dir` (the late chain) unconditionally ignores its cache
whenever `--reimport` is set - a deliberate "refresh from the source" rule -
so an early check relying on that same cache to decide "already handled"
just watched the late chain re-ask anyway, mid-cook. Fixed by keying the
early check's cache-skip to the same flag the late chain uses.

**Bug 2 (this branch).** `--reimport PATH` resolves a fresh source in a
different code path from the questionnaire's map slot, and never called
`_sumo_for_local_pick` at all - so a *replayed* setup (`--profile X
--reimport`) never got the early ask; only the interactive row-2-reopen case
did. Fixed by calling it at both new points `main()` can establish a source.
Bug 1 is a prerequisite - without it, calling the helper here would always
no-op, since `args.reimport` is unconditionally true at both call sites.

Guarded against double-asking: an explicit path always wins over (and
re-asks past) whatever a fresh interactive pick already asked about; the
reopened-dialog branch only fires when no fresh interactive pick was made.

Verified against the full flag matrix (cache present/absent × reimport
true/false × explicit `--sumocfg` × app with a launcher): the previously
broken case now asks; every other case is unchanged.

PR #363, opened 2026-09-17
setup: two messages that named the wrong culprit

yunlishao-vibe:
Both of these are messages that are true and point at the wrong thing. Neither
changes behaviour on a working machine.

## 1. A comment about a BOM, containing a BOM

`Carla/run_cosim.py` explains the `FIXS_VERSION.txt` BOM bug in a comment — and
the comment contains a literal BOM, the character being described, pasted into
the prose at lines 130 and 132. Python 3 reads it without complaint, so it sat
there harmlessly:

```
printf 'x = 1\n# BOM survives as \xef\xbb\xbf here\nprint(x)\n' > bomtest.py
python bomtest.py     ->  1   (exit 0, CPython 3.10.11)
```

Until a user's PATH `python` turned out to be a 2.x, whose tokenizer stops at the
first non-ASCII byte:

```
  File "D:\ORNL\FIXS_Application\beta\FIXS\Carla\run_cosim.py", line 130
SyntaxError: Non-ASCII character '\xef' in file ...\run_cosim.py on line 130,
but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

That wording is python 2's tokenizer — python 3 has no such message. So the
report names a line inside the build, never mentions the interpreter, and reads
as a broken release. (Their bundle was `3a70dab`, this branch's base.)

The launcher-side fix — refusing to hand the engine to a python 2 at all —
belongs in FIXS_Applications and is
[there](https://github.com/ORNL-Real-Sim/FIXS_Applications/pull/65). This PR just
takes the two bytes out and writes `U+FEFF` in words, which is what the comment
meant. `Carla/run_cosim.py` is now pure ASCII. Had the BOM not been there, python
2 would have died a few lines later on the first of the file's 471 f-strings — so
this is not a python-2 compatibility fix and is not meant as one.

## 2. `pip install carla==0.9.15 failed.`

On an interpreter outside CPython 3.7-3.10 there is no carla wheel and no source
distribution, so pip reports only that it found no matching distribution, and
`ensure_carla` relays a bare failure that reads as a network or index problem.
Nothing in `carla_env_setup.py` checks the interpreter's minor version.

This matters on a machine without conda, where setup pip-installs into whatever
interpreter it is running under — increasingly a 3.11+ on a fresh install.

`_carla_wheel_hint()` names the version when the version is the reason, and
returns `""` when it is not, so it never talks over a real failure:

```
[setup] pip install carla==0.9.15 failed.
[setup] that interpreter is python 3.12. The CARLA client is published
        only for CPython 3.7-3.10 and has no source distribution, so
        there is no wheel for it to install - this is not a network
        problem. Bind a 3.10 env instead:
            conda env create -n realsim -f environment.yml
            python carla_env_setup.py --update-python
```

Exercised against `_python_tag` returning each of: a real 3.10 interpreter (`""`),
`cp312` (the message above), `cp37` (`""`, in range), `None` (`""`), and a
malformed `cp3x` (`""`).

## Tests

`tests/Sumo/Carla/test_carla_env.py` — 53 passed, 18 failed. All 18 failures are
pre-existing and in a file this PR does not touch: at the base commit,
`Carla/import_map.py:3089` defines `parse_selection` while the test calls
`import_map._parse_selection`. Worth a separate fix; flagged, not bundled here.


yunlishao-vibe:
Two more commits, on the same theme: `client` mode means **no CARLA on this machine**, and nothing more.

### The blocker

A machine that answered [3] could not run the one thing it is obviously for:

```
run_cosim.py --profile roosevelt_default --sumo-only
[cosim] carla.json is 'client' mode (no CARLA on this machine), but
        roosevelt_full_config_fast.yaml points CarlaSetup.CarlaServerIP at
        127.0.0.1, which IS this machine.
exit=1
```

Stock yamls point `CarlaServerIP` at 127.0.0.1, so that check fired on every single-machine traffic-only run, 130 lines before the `--sumo-only` path.

Nothing about a remote CARLA needs the mode to say so. `--peer HOST[:PORT]` writes `CarlaSetup.CarlaServerIP/Port` and implies `--no-launch`; `--carla-host` is the same at the RPC level. Both are decided from `_is_local_host(args.carla_host)`, neither consults `mode`. And the tree already reads `client` as "no local CARLA" — `load_config` calls it "the one mode that is complete without a carla_root", and `import_map` sets `sumo_only = _mode() == "client"`, skipping 368 MB of roosevelt_full's 370.

### Changes

Menu [3] now says what it means. The CarlaServerIP check and the who-must-cook advisory stand down for `--sumo-only`. `--sumo-only` implies `--no-launch`, which is what stops the source-build cook, the bundle's carla half and TL/sign placement — all three were gated on `--no-launch` alone, so a traffic-only run on a source box still did CARLA-local work, including an Unreal cook on a fresh map. The carla client install is offered rather than required; declining used to `sys.exit`.

The run-setup summary no longer prints `CARLA  client (none on this machine)  ->  127.0.0.1:2000  [this machine]`, a row that contradicts itself and opens onto nothing. `show()` and the reader both build from `visible_slots()`, because the printed numbers are the edit keys (`SLOT_KEYS[n-1]`) and renumbering one without the other is how "5" opens SUMO while the screen says CARLA.

### Verified

Same probe that exited 1 now completes: `[cosim] SUMO-only stack up: SUMO + TrafficLayer, no CARLA`, exit 0, no processes left behind. Menu rendering checked from the branch. Row hiding checked with two controls that must keep the row — a remote endpoint, and `source` mode — both keep it with "5" still opening CARLA, so the check is not vacuous.

Not changed, worth a look separately: the engine row still reads `cpp (TrafficLayer + VirCarlaEnv, from the yaml)` on a no-CARLA box, and VirCarlaEnv is not started for `--sumo-only`.

yunlishao-vibe:
Two more commits. The row hiding in the previous one keyed on the MACHINE having no CARLA (`mode == "client"`), which missed the commoner case: `--sumo-only` on a full source build.

`derived_from_yaml` already took `args`, so it now reports what this RUN will do with what the yaml says, not only what the yaml says. On the reported setup:

```
[cosim] --sumo-only: no CARLA is launched, loaded or dialled; implying --no-launch.
[cosim] Run setup 'mlk_eco_driving_l0_warmup':
   1) app       mlk_eco_driving
   2) map       mlk_no_signal              (Digital-Twin-Library)
   3) scenario  config_Sumo_ecoDriving.yaml
   4) engine    py                         (TrafficLayer only - --sumo-only starts no bridge)
   5) SUMO      gui, as fast as possible
```

Two further rows were describing CARLA work that a traffic-only run does not do. The engine row named the bridge from the yaml (`TrafficLayer + mainVirCarla.py`) when `--sumo-only` starts no bridge — annotated rather than hidden, because TrafficLayer genuinely runs and the value is what the next CARLA run uses. The SUMO row reported `CARLA tick 0.1 s`, the rate CARLA is stepped at, on a run with no CARLA to step.

Controls that must keep the CARLA row, all still keeping it: an ordinary run on the same source box, a `--peer` remote endpoint, and a no-CARLA box driving a remote CARLA.

yunlishao (approved):

PR #366, opened 2026-09-21
An app's controller is a .py, not a wrapper that calls one

yunlishao-vibe:
## What

`launch` in `apps.json` now accepts a controller's `.py` directly and runs it under
`sys.executable`. An application no longer needs a wrapper script.

```jsonc
"launch": "my_controller.py --penetrationRate 0.0"
```

## Why the wrapper was never a choice

Two platform facts made it mandatory:

- `subprocess.Popen(['my_controller.py'])` fails on Windows with `WinError 193` —
  `CreateProcess` does not consult `PATHEXT`.
- `launch: "python my_controller.py"` resolves only the **first** token, so
  `launch_command` looks for a `python.bat` in the app folder, does not find it,
  warns, and starts nothing.

So every app shipped the same file: resolve a directory, call python on the `.py`
beside it, propagate the exit code. `mlk_eco_driving`'s is 97 lines — 58 comments,
9 blank, 30 statements — and exactly one block of it is that app's own (turning
`FIXS_SUMOCFG` into its controller's `--sumoFolder` / `--sumoConfig`).

`sys.executable` needs no new plumbing: it is the interpreter `run_cosim` re-exec'd
into and applied the app's `requirements.txt` to, and the same one it already passes
as `FIXS_PYTHON`. That is also why a wrapper going looking for its own interpreter
could find an env that never received those requirements.

## The handoff is untouched

`FIXS_HANDOFF` travels in the `env` dict the caller hands `Popen`, never on this
argv, so it reaches a python child exactly as it reached a `cmd.exe` one. Verified
end to end by replicating `start_app`'s spawn-and-poll with no wrapper in the chain:
the child read the variable, wrote `{"sumocfg": ...}` atomically, and the parent
read it back.

A wrapper stays right where it has something of its own to carry — arguments
computed at launch time, an environment to set, a controller that is not python —
and an extensionless `launch` still resolves to `.bat` / `.sh` as before.

## Also: three docstrings sent developers to a file that does not exist

`controller_template.py` and `EgoControllerHost.py` documented
`EgoActuationSource:` / `EgoController:` at the top level. No parser reads those —
`ConfigHelper` reads `EgoSetup.ActuationSource` and `EgoSetup.Controller` — and the
template pointed at `CommonLib/VirEnv/IEgoController.py`, which is not in the tree.
Someone copying the shipped template was being sent to a missing file to learn keys
that do nothing.

## Tests

`tests/Python/unit/test_launch_command_python.py`, 7 cases. The load-bearing one is
not the shape of the argv but that it **spawns** — and `test_bare_py_is_what_fails`
is its negative control, asserting the un-prefixed form is the `WinError 193` this
removes.

```
7 passed in 0.19s
```

Negative-controlled against the unfixed `launch_command` — 3 of the 7 fail there,
with exactly that error:

```
FAILED test_py_is_run_under_this_interpreter
FAILED test_py_launch_actually_spawns
FAILED test_exit_code_reaches_the_caller
3 failed, 4 passed
```

Full unit suite: `1 failed, 333 passed, 8 skipped`. The failure
(`test_fixs_api.py::test_several_subscriptions_require_a_port`,
`'_Cfg' object has no attribute 'raw'`) reproduces identically on pristine
`origin/dev_v0.9.0`, so it is not from this branch.

## Not fixed here, found while checking

Two pre-existing failures on `dev_v0.9.0`, both confirmed against a clean checkout
of the base and both unrelated to this change:

1. `test_fixs_api.py::test_several_subscriptions_require_a_port` — `_Cfg` stub in
   the test lacks `.raw`, which `_selectSubscription` now reads.
2. `tests/VirEnv/test_ego_controller.py::test_shipped_template_loads_and_commands` —
   the template's `setup()` calls `fixs.xil.enabled()` (#357), which raises when
   `$FIXS_CONFIG_YAML` is unset, and the test does not set it.


yunlishao-vibe:
Verified end to end against a real stack, not just the unit tests.

The FIXS bundle in a FIXS_Applications checkout was fetched from `b60c4dd` — this
branch's base — and its `Carla/app_catalog.py` was byte-identical to pristine
`origin/dev_v0.9.0`. Patching that one file with this change is therefore exactly
what the release will ship. `mlk_eco_driving` then had both launcher scripts deleted
and its `apps.json` pointed straight at its controller:

```jsonc
"launch": "eco_driving_mpr_SUMO_MLK.py --penetrationRate 0.0"
```

`run_cosim --app mlk_eco_driving --sumo-only`:

```
[APP]  eco_driving_mpr_SUMO_MLK.py --penetrationRate 0.0
[cosim] 'mlk_eco_driving' builds its scenario from the map's: ...\mlk_notexture_uturn.sumocfg
[fixs.sumo] reported scenario -> ...\MPR\..._config_Sumo_ecoDriving\mlk_notexture_uturn_ego.sumocfg
[cosim] SUMO scenario: 'mlk_eco_driving' generated ...
[cosim]   OK   SUMO TraCI listening on 1337
[cosim]   OK   TrafficLayer serving the bridge on 430
[fixs] connected to TrafficLayer on 127.0.0.1:430
Connected to FIXS server at 127.0.0.1:430
Starting subscription
```

followed by 6200 lines of controller output with the ego present, moving, and
tracking signal phase (`sim_time` advancing, `next_tls_id 202639124`,
`curr_status green`, `dist2Stop_ft` falling). Ctrl-C teardown left no orphaned
`TrafficLayer.exe` or `sumo-gui.exe`.

The part worth calling out: **`FIXS_HANDOFF` needed nothing.** It travels in the
`env` dict `start_app` hands `Popen`, never on the argv this function builds, so it
reached a python child exactly as it reached `cmd.exe`. The scenario handshake is
unchanged.

The companion PR is ORNL-Real-Sim/FIXS_Applications#69, which deletes both MLK
wrappers. It cannot merge until a release carries this change.

yunlishao (approved):

Issue #373, opened 2026-09-22
[BUG] the first exchange after a warm-up costs ~2 s in one tick


yunlishao:
**Describe the bug**

With `SimulationSetup.WarmUpUntilEgoEntry: true` and `WarmUpServePorts: [430]`, the CARLA bridge on port 440 sits out the whole warm-up, so nothing asks CARLA anything for a minute of wall clock. When the boundary opens, the co-simulation freezes for about **2 s inside the two ticks around ego entry**, then runs normally — no tick over 78 ms for the rest of the run.

It is not a stall in the usual sense, which is why it is easy to miss: the log keeps printing ticks, and only the wall clock between them shows it. Direct tic-toc inside the bridge loop, stock `dev_v0.9.0` code (the bundle measured is byte-identical to `dev_v0.9.0` HEAD for every file involved):

```
[tic] SLOW TICK simTime=0.10  total 63.047 s      <- first exchange carrying traffic
[tic]     core.runStep(total)                  62.469 s
[tic]       of which: fixs recv (waiting)      61.969 s    blocked on the warm-up, not work
[tic]       of which: spawn/despawn/stage       0.500 s    197 vehicles, all in one exchange
[tic]     z audit                               0.578 s    its first world.get_map()

[tic] SLOW TICK simTime=0.15  total  0.922 s      <- first controlled tick
[tic]     runController                         0.906 s
[tic]       BehaviorAgent(...)                  0.734 s    ~0.58 s of it world.get_map()
[tic]       carla.bind(route=True)              0.157 s    route snap, 2561 waypoints
```

Three causes, all the same shape — work whose *first* use is deferred into the loop because the warm-up gives it no earlier occasion:

1. **`carla.World.get_map()` is paid twice, both times inside the entry ticks.** It serialises and re-parses the whole OpenDRIVE and is not shared between callers: called twice back to back in one process it measured **0.640 s** and **0.578 s**. `CarlaBackend.auditZAlignment` pays one (`self._map = self._world.get_map()` on its first call, which is the first exchange that carries vehicles); CARLA's own `BasicAgent.__init__` pays the other on the first controlled tick. Neither can happen during the warm-up as things stand, because no traffic arrives and no ego exists.

[...]

PR #374, opened 2026-09-22
cosim: the warm-up boundary exchange, and two defects it exposed

yunlishao-vibe:
Closes #373.

## Summary

With `WarmUpUntilEgoEntry` the CARLA boundary is shut for the whole warm-up, so nothing asks CARLA anything for a minute. Three pieces of *first-use* work then land in the two ticks where the ego enters. Measured by tic-toc inside the bridge loop, stock `dev_v0.9.0`:

```
SLOW TICK simTime=0.10  total 63.047 s      <- first exchange carrying traffic
    core.runStep(total)                  62.469 s
      of which: fixs recv (waiting)      61.969 s    blocked on the warm-up
      of which: spawn/despawn/stage       0.500 s    197 vehicles in one exchange
    z audit                               0.578 s    its first world.get_map()

SLOW TICK simTime=0.15  total  0.922 s      <- first controlled tick
    runController                         0.906 s
      BehaviorAgent(...)                  0.734 s    ~0.58 s of it world.get_map()
      carla.bind(route=True)              0.157 s    route snap, 2561 waypoints
```

Nothing after that exceeds 78 ms. It is a transient, and it lands exactly where a bench in the loop is first being commanded.

Chasing it turned up **two defects that predate the pool**, and they are the more important half of this PR.

## The timing work

**`world.get_map()` is paid at start-up.** It serialises and re-parses the whole OpenDRIVE and is not shared between callers -- 0.640 s and 0.578 s when called twice back to back in one process. `initTrafficPool` now takes it along with the blueprint library; the core calls that on the first step, the one iteration that does no recv, so it runs in the warm-up dead time.

**`fixs.carla.precompute_route()`** lets a controller build the route plan in its constructor. The plan is EgoRoutePoints snapped to the map; neither waits for the ego, only `bind` does.

**`CarlaSetup.SpareVehiclePool`** parks actors up front and hands them out as the traffic arrives. `try_spawn_actor` measured 1.6-3.8 ms each, so ~200 arrivals in one exchange is half a second. Default 0, which is today's behaviour exactly. Leftovers are trimmed one exchange after the pool is first drawn from, because a parked actor costs ~0.03 ms of *every* `world.tick` just by existing -- measured, and the same whether it sits below the map or 20 km away, so it is per-actor bookkeeping and not rendering.

## The two defects

Turning the pool on moved the ego 150 m off its baseline. It was not float noise, and the pool was not really at fault -- it exposed these.

**The spawn height.** CARLA refuses a spawn whose bounding box intersects the road, and the larger blueprints do at `_kSpawnOffsetZ = 0.1`. On the MLK warm-up burst, 9 of 197 vehicles were refused -- eight `vehicle.mercedes.coupe`, one `vehicle.bmw.grandtourer`. Each was retried **alone in an empty world at the same transform** and refused again, while a `vehicle.tesla.model3` at that identical transform was accepted:

```
id        its blueprint          alone?       as a tesla.model3?
14.14     mercedes.coupe         REFUSED      placed
21.34     mercedes.coupe         REFUSED      placed
7.17      bmw.grandtourer        REFUSED      placed
```

So it is the model, not the spot and not congestion. Sweeping the lift: 0.1 -> 9 refused, 0.2 and above -> 0 refused. Now 0.5, with margin, and free -- the spawn pose is transient, overwritten by `setVehiclePose` in the same tick before `world.tick`. A refused vehicle is skipped and retried next exchange, so the corridor silently lost nine cars for 0.1 s at every burst. Peak mirrored vehicles went from 199 to **215** once they stopped being dropped.

**The blueprint draw.** It took the next value from one shared seeded generator, so a vehicle's model depended on how many had drawn *before* it -- and a refused spawn consumes a draw, then draws again on the retry. Nine refusals shifted the sequence by nine and re-dealt every later vehicle: **46 of 241** mirrored vehicles got a different model. Not cosmetic -- the blueprint fixes `bounding_box.extent.x`, which is the pose anchor, so those 46 sat up to **1.58 m** from where they had been, permanently, in front of an agent reading them as its leader:

```
id        ticks  min m       max m       spread      verdict
6.36      228    1.575036    1.575149    0.000112    CONSTANT -> blueprint
23.52     398    1.409532    1.410286    0.000754    CONSTANT -> blueprint
21.53     388    0.236557    3.209264    2.972707    varying  -> timing
```

`map_Sumo_vClass_to_Carla_blueprintId` now takes the vehicle id and derives the choice from `(seed, id)` with blake2b -- not `hash()`, which Python randomises per process and which would have made the seed useless.

## Result

With both defects fixed, **the pool is free**:

```
pool 220 vs pool 0, 3000 exchanges of MLK eco-driving L2
  max |position| difference 0.000000 m   IDENTICAL on every row
```

Before them, the same comparison diverged at the 130th row and reached 150 m.

And the entry cost it was all for:

| | worst feed in the run | feeds over 0.25 s |
|---|---|---|
| stock | 1.100 s | 2, totalling 2.07 s |
| this branch | **0.110 s** | none |

0.110 s against a 0.1 s feed period -- the entry tick is no longer distinguishable from an ordinary one.

## This moves existing baselines, deliberately

The old baseline had nine cars missing and 46 wearing the wrong body. Ego speed over 300 s against `rung3_user_0916`:

```
mean speed         5.4065 -> 5.3337 m/s
distance           1621.9 -> 1600.1 m   (-1.35%)
mean |difference|  0.76 m/s
p95  |difference|  3.09 m/s
max  |difference|  6.14 m/s   where a signal is caught differently
```

Aggregates are within 1.4%; individual moments are not, because a metre of difference in a leader's position decides whether the ego clears a green. Baselines need regenerating once, after which the ladder is reproducible **and** the pool costs nothing.

## Verification

23 new unit tests. Because most assert that something stays the same, each was checked against a mutation rather than trusted:

| mutation | tests failed |
|---|---|
| pool never hands anything out | 3 |
| map prefetch dropped from `initTrafficPool` | 1 |
| trim destroys handed-out vehicles too | 2 |
| route cache never consulted / never filled | 1 / 1 |
| revert to the sequential blueprint draw | 4 |
| use `hash()` instead of blake2b | 1 |
| ignore the blueprint seed | 2 |

Full unit suite: **1 failed, 398 passed, 8 skipped**. The failure is `test_fixs_api.py::test_several_subscriptions_require_a_port`, confirmed present on the base commit `b2089c74` in a clean worktree before this branch existed.

The route precompute safety was measured, not assumed: seeding the snap at the ego z (205.29) against the no-ego fallback (210.21) gave **0 of 2560 differing road/lane**, geometry identical to 0.0000 m -- the seed feeds only the first query and the elevation is carried forward thereafter. That comparison is not blind: holding z at 0 for every query differs at **341 of 2560**.

## Checklist
- [x] Tests pass locally
- [x] Documentation is updated (config key and both constants carry their measurements)
- [x] Issue linked above

## Additional Notes

**Pool sizing is per scenario and manual.** 220 suits MLK because its peak concurrency is 216 (`running=` in SUMO's `simulation_summary.xml`). The bridge reports the peak it saw at shutdown so the number is never a guess. Learning it automatically was considered and dropped: writing it back into the staged yaml is exactly what #372 stopped.

**Python bridge only.** `VirCarlaEnv.exe` ignores `SpareVehiclePool`; the blueprint and spawn-height fixes have C++ peers that need the same treatment for the two bridges to agree.

**Blueprint matching is not perfect by construction.** 20 of 196 vehicles in the verification run found no spare of the model they drew and spawned normally -- the price of leaving each vehicle model alone. It costs ~30 ms rather than ~500.

**Separate trap worth its own issue.** A run that dies during start-up still gets its half-resolved state saved into `run_profiles.json`, so the *next* run fails with an unrelated "config not found". Hit twice while producing these measurements.


yunlishao-vibe:
Two findings from review that belong on the record here.

**The spare pool in this PR duplicates one CarMaker already has.** `CarMakerBackend::spawnVehicle` pops from `carQueue_`/`truckQueue_` and `despawnVehicle` parks the slot at z = -5000 and pushes it back — a free list with recycling, which is what this PR reimplements for CARLA without the recycling. Filed as #379, with the constraint that CarMaker's pool is mandatory and fixed while CARLA's must stay optional (`SpareVehiclePool: 0` = create on demand, the behaviour every existing result was produced with).

**`parkSpares()` and `SpareVehiclePool` are now two unrelated meanings of "spare" in one interface** — a per-refresh drift correction for CarMaker's FreeMotion slots (#168) and a start-up pool for CARLA. #379 proposes keeping the verb CarMaker-only to resolve that.

Also measured while answering a review question, in case it saves someone the experiment: CARLA cannot be made to create an actor where it refuses to. `try_spawn_actor` returns None, `spawn_actor` raises and `carla.command.SpawnActor` returns an error, all with the same server-side reason (`Spawn failed because of collision at spawn position`). There is no override flag, which is why the fix here is to ask for a transform that clears the check rather than to bypass it.

yunlishao (approved):

